PR #115, opened 2013-09-15
Allow custom authenticator to be added via configuration

georgy:
Introduce new configuration parameter (authenticatorClass) that is used to specify custom authenticator class. This authenticator class will be found, instance will be created (if class has a constructor that takes Configuration as parameter, jolokia runtime configuration will be provided, allowing authenticator to take it into account). Newly created instance of authenticator will be set for Jolokia server to use. If authenticatorClass config parameter was not provided, code will fall back to old logic (i.e. check if user/password is set and create default authenticator).

This pull request also adds a ugly looking code into copyResourceToTemp method of JvmAgentConfigTest with the main goal of properly escaping paths used in tests when they are run win (tested on win7).

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -24,6 +24,12 @@
   </properties>
   <body>
     <release version="1.1.4-SNAPSHOT" description="Release 1.1.4-SNAPSHOT">
+      <action dev="georgy" type="add" date="2013-09-25">
+        Introduced new configuration parameter "authenticatorClass" which can be used to provide custom authenticator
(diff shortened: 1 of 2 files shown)


rhuss:
FYI, I'm about to release 1.3.1 today and for consistencies sake I renamed 'authenticationClass' to 'authClass' as parameter. I hope this does not provide you any issues when doing upgrades. If this is a stumbling block, though, I can introduce this parameter back as a fallback (but would like to avoid this to keep the code base clean).

Issue #125, opened 2014-01-19
add a discovery mechanism to jolokia so its easy to discover JVMs (e.g. via local files, zeroconf/bonjour etc)


jstrachan:
it'd be nice to support zeroconf to advertise the URLs to connect to jolokia on a machine using this library:
http://jmdns.sourceforge.net/

then folks could discover and connect to jolokia based JVMs more easily; e.g. there are pure C / ObjectiveC / JavaScript clients of zeroconf available

PR #126, opened 2014-01-23
first spike of #125 to add a discovery mechanism to jolokia; which defau...

jstrachan:
...lts to using files on the local machine (with keep-alive and tidying up of old files), a little JMX API and an experimental zeroconf plugin implementation.

currently only the JVM agent is supported; the challenge for the servlet/osgi agents is figuring out the jolokia URL before the servlet's been invoked...


davidkarlsen:
I don't see this filebased discovery in the codebase any longer? Was it removed in favour of the multicast based solution? It would be a nice way to provide discovery when using jolokia in hawtio when deployed as a central .war

Issue #178, opened 2015-01-08
list operation causes InstanceNotFoundException on JBoss for ResourceAdapterModule


jcordes73:
When installing jolokia on JBoss EAP 6.3 with a resource-adapter configured (f.e. for JBoss A-MQ) and calling the list operation an **InstanceNotFoundException** is generated:

"javax.management.InstanceNotFoundException: JBAS019906: Could not find ResourceAdapterModule"

The full response below:

[...]

PR #179, opened 2015-01-08
Fix for https://github.com/rhuss/jolokia/issues/178.

jcordes73:
I have added a special handling of InstanceNotFoundException in order to solve https://github.com/rhuss/jolokia/issues/178


rhuss:
Thanks a lot. Really strange that JBoss can not find its own MBeans which it returns after a query, but well. That's life ;-)

I assume that APL is fine for your contribution (as stated [here](https://github.com/rhuss/jolokia#contributions))

PR #183, opened 2015-01-31
StringToObjectConstructor now can create object by its constructor.

khaing211:
Signed-off-by: Khai Nguyen <khaing211@gmail.com>

### DIFF
--- a/agent/core/src/main/java/org/jolokia/converter/object/StringToObjectConverter.java
+++ b/agent/core/src/main/java/org/jolokia/converter/object/StringToObjectConverter.java
@@ -171,22 +171,26 @@ public class StringToObjectConverter {
         Class givenClass = pArgument.getClass();
         if (expectedClass.isArray() && List.class.isAssignableFrom(givenClass)) {
             return convertListToArray(expectedClass, (List) pArgument);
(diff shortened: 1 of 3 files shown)


rhuss:
Thanks a lot for your pull request, I think something like that would be a quite addition for upstream serialization. 

However there are some issues:

* A minor one, `Constructor.getParameterCount()` is only available since Java 8 but Jolokia still supports Java 7. The fix is trivial, though.
* More seriously, the change is not backwards compatible. E.g. for `Date` we a have a own extractor `DateExtractor` which converts strings to dates. However, `Date` has also a single string argument constructor, so your patch would pick that one first which results in a complete different behavior for converting string to date. Maybe we can move the constructor based conversion to the very end of the chain, so that it least its backwards compatible. 

I can have a look at the last issue, too, but I'm currently quite time constrained. Maybe you can fix that ? (e.g. I believe the logic should go into `convertFromString()` at the end).

Also, could you please update the documentation (in `protocol.xml`, section *serialization-request*) ? That would be awesome ....

khaing211:
Thank for your review. I updated per your suggestion.

rhuss:
Thanks a lot ! 

It will be available in the next release.

PR #194, opened 2015-05-06
Add String to AttributeList converter

hughjfchen:
Add a String to javax.management.AttributeList parser which can parse a sring and construct the appropriate AttributeList object. 

The string must be in following format:
    attribute-name1=attribute-type1:attribute-value1,attribute-name2=attribute-type2:attribute-value2...
The attribute types could be as following:
I: int
L: long
F: float
D: double
B: boolean
T: short
S: String
Y: byte
C: char
If there're sepcial characters within the attribute value, the following escape rules applied:
    1.for "," and "=",put three PATH_ESCAPE characters before it
    2.for ":",put four PATH_ESCAPE characters before it

Please focus on the agent/core/src/main/java/org/jolokia/converter/object/StringToObjectConverter.java file and ignore the other websphere specific files.
Thanks,


rhuss:
Thanks for the PR ! Will have a look asap, but probably not before 1.3.2

tadayosi:

PR #225, opened 2015-10-28
Add authClass configuration option to jolokia-osgi

ryandgoulding:
jolokia-jvm has the option to specify an authClass.  When specified, authMode
is ignored and authentication responsibility is delegated to authClass.  This
change enables the use of any authClass which extends
org.jolokia.osgi.security.Authenticator.

Signed-off-by: Ryan Goulding <ryandgoulding@gmail.com>

### DIFF
--- a/agent/osgi/src/main/java/org/jolokia/osgi/JolokiaActivator.java
+++ b/agent/osgi/src/main/java/org/jolokia/osgi/JolokiaActivator.java
@@ -1,6 +1,8 @@
 package org.jolokia.osgi;
 
 import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.util.Dictionary;
 import java.util.Hashtable;
 
@@ -202,7 +204,7 @@ public class JolokiaActivator implements BundleActivator, JolokiaContext {
             if (config == null) {
                 return null;
             }
-            Dictionary props = config.getProperties();
+            Dictionary<?, ?> props = config.getProperties();
             if (props == null) {
                 return null;
             }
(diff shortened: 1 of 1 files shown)


ryandgoulding:
@rhuss Please let me know your thoughts on this change.  Glad to make any modifications you suggest.  I am trying to integrate this with OpenDaylight (osgi/karaf based);  we have an existing AAA subsystem and I am trying to delegate jolokia authN responsibility to that.  I am the Project Technical Leader of AAA in ODL, so let me know if you think there is a better way to do this and I can adapt internally.

rhuss:
Thanks a lot for the PR. In fact an extension point for an restrictor makes much sense. Jolokia 2.0 already has this, but it's still in early development stage. 

Since I'm currently on vacations I will take care about this PR after this (2 weeks). Hope this is OK.

ryandgoulding:
Thanks @rhuss and enjoy your vacation!

rhuss:
Sorry for the long delay. However, I'm just about to pick this up now and cut a release before the end of this week, 

Would an OSGi Authenticator service help you, too ? I ask because in an OSGi environment this is the canonical way to wire stuff together and we do this already for restrictors which are exposed as OSGi service. That way you could also nicely stack multiple authenticator services (much like for the restrictor). 

My suggestion is to use a new `authMode` with value `service` in which case in an OSGi context the authentication service is looked at the time of a request. If no such service is available, then the access is restricted. If more than one service is available then all services must pass. An extension could be, to add an extra `authRequire` option with possible values `all` or `any` and allow also multiple, comma separated `authMode` values. 

I think that would fit better in the current implementation (i.e. avoiding `Class.forName()` lookups in an OSGi environment).

What do you think ?

rhuss:
Seel also http://njbartlett.name/2010/08/30/osgi-readiness-loading-classes.html why using `Class.forName()` is not the way to go in OSGi ...

ryandgoulding:
Good points, rhuss.  I will look into adding a service interface instead of using Class.forName(...).

Issue #224, opened 2015-10-28
Custom restrictor using JolokiaHttpHandler


pwielgolaski:
I wonder what are my options to use custom restrictor when I have already HttpServer setup (sun).
I would like to define my own restrictor, not using policy.

Could we?
- make protected method Restrictor createRestrictor(Configuration pConfig)
- give possibility to define class for restrictor in configuration
- accept restrictor in constructor in similar way as AgentServlet does it

rhuss:
Yes, restrictor should be more extensible (Jolokia 2.0 already will provide the proper hooks)

Since I'm currently on vacations this might take two Wood until I can pick that up. Hope that's OK.

nevenr:
Hi guys,

if you think this change (https://github.com/nevenr/jolokia/commit/94a1ed5f6fa8aa28dc34b84d17e26517b48db788) is good enough I will create PR.

Regards,
N.

grgrzybek:
Closed as fixed here: 94a1ed5f6fa8aa28dc34b84d17e26517b48db788

PR #228, opened 2015-11-17
Custom restrictor using jolokia http handler #224

nevenr:


rhuss:
Happy new yeart ! ;-)

Thanks a lot for the PR and sorry for the long delay. But finally I included all your changes, with some slight adoption.

Mostly I moved "--encrypt" to a dedicated "encrypt" command for encrypting the password (and changed the marker code to "[[" ... "]]" for consistencies sake with jmx4perl encryption).

PR #241, opened 2016-01-27
Fix version range in jolokia-agent-parent pom.xml

tokuhirom:
We want to change as expected of implementation to the specifications.
https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html.

### DIFF
--- a/agent/pom.xml
+++ b/agent/pom.xml
@@ -53,7 +53,7 @@
         <module>mule</module>
       </modules>
       <activation>
-        <jdk>[1.7,</jdk>
+        <jdk>[1.7,)</jdk>
       </activation>
     </profile>
   </profiles>


tokuhirom:
I got a same error reported for gson. ref. https://github.com/google/gson/issues/596

rhuss:
Thanks for spotting this !

krrrr38:
Hi, @rhuss :)

In my situation, I got following error to get jolokia-core 1.3.2 dependency.
```
Invalid JDK version in profile 'mule': Unbounded range: [1.7, for project
```
It might be fixed by this feature.

Is there any SNAPSHOT version which includes this feature?
and do you have plan to release next version?

rhuss:
I plan to make a 1.3.3 release soon (probably at the beginning of next week).

In the meantime I just pushed a 1.3.3-SNAPSHOT so please let me know, whether this works for you.

krrrr38:
I will use 1.3.3. Thank you for quick response!

PR #243, opened 2016-02-08
Implement streaming json responses

gnufied:
This pull request implements streaming json responses. Which user can configure via jolokia properties file. 

Using streaming json I saw a drop by 39 times - in allocation sizes while an external python script performed 10 identical collections on the url. I did this numerous times and while numbers vary somewhat, it still holds true that using streaming json results in less allocations.

~~Should we choose to merge this, I will be happy to get tests sorted for the same.~~ I am also sure, I have only covered jvm-agent output and this feature probably will not work for other use cases. 

gc-profile1
gc-profile2

### DIFF
--- a/agent/jvm/src/main/java/org/jolokia/jvmagent/handler/JolokiaHttpHandler.java
+++ b/agent/jvm/src/main/java/org/jolokia/jvmagent/handler/JolokiaHttpHandler.java
@@ -32,7 +32,6 @@ import javax.management.RuntimeMBeanException;
 import javax.security.auth.Subject;
 
 import com.sun.net.httpserver.*;
-import com.sun.org.apache.xml.internal.serializer.OutputPropertiesFactory;
 import org.jolokia.backend.BackendManager;
 import org.jolokia.config.ConfigKey;
 import org.jolokia.config.Configuration;
@@ -44,7 +43,6 @@ import org.jolokia.restrictor.*;
 import org.jolokia.util.*;
 import org.json.simple.JSONAware;
 import org.json.simple.JSONStreamAware;
-import sun.net.www.http.ChunkedOutputStream;
 
 /**
  * HttpHandler for handling a Jolokia request
--- a/agent/jvm/src/main/java/org/jolokia/jvmagent/handler/JolokiaHttpHandler.java
+++ b/agent/jvm/src/main/java/org/jolokia/jvmagent/handler/JolokiaHttpHandler.java
@@ -32,6 +32,7 @@ import javax.management.RuntimeMBeanException;
 import javax.security.auth.Subject;
 
 import com.sun.net.httpserver.*;
+import com.sun.org.apache.xml.internal.serializer.OutputPropertiesFactory;
 import org.jolokia.backend.BackendManager;
 import org.jolokia.config.ConfigKey;
 import org.jolokia.config.Configuration;
@@ -42,6 +43,8 @@ import org.jolokia.jvmagent.ParsedUri;
 import org.jolokia.restrictor.*;
 import org.jolokia.util.*;
 import org.json.simple.JSONAware;
+import org.json.simple.JSONStreamAware;
+import sun.net.www.http.ChunkedOutputStream;
 
 /**
  * HttpHandler for handling a Jolokia request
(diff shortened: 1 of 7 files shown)


rhuss:
Thans a lot, and sorry for the delay. 'was skiing in the Alpes for a week ;) 

I will have a look at it, and will include it in the one way or other. Could be that it will be added to Jolokia 2.0, but let me have a look first. This might take until the end of the week, though ....

gnufied:
Cheers. thanks for having a look.  My motivation is to use this agent with cassandra cluster which uses ParNew GC and hence can result in stop-the-world gc pauses on young gc collection.  I have profiled extensively with cassandra server and these changes have made young GC collections almost disappear.

The reason, we see less allocations is, there is lot of wasteful string creation when serializing full json strings. for example,  when `JSONObject` serializes the `Map` and the key is a string, it undergoes following transformation:

```
String -> char array (because I can only write char array to Writer interface)
Char array -> String (because we are asking full json string back via JSONString which uses StringWriter)
String -> byte array (because final output is written to outputput stream )
```

This code removes lot of these redundant allocations. I haven't looked too closely at other endpoints (such as when ran within a container or proxy mode), since my own usecase  uses jvm agent mode prominently. But I am happy to look. Also, do you think, it should be configured on per-request basis too?

rhuss:
was finally able to have a look, thanks a lot for the PR.

I wonder what would be the drawback to exclusively use streaming (and get rid of the old code). If streaming is always superior, we could get rid of the extra option and make that the default and only behaviour.

Do you see any problems in always using streaming ?

gnufied:
At Yelp - we have been running modified version of Jolokia (with this patch enabled) on hundreds of machines and so far we haven't seen a problem. 

My initial worry was - what if a particular HTTP client couldn't deal with chunked encoding and this patch broke some of the tooling we have. But after good amount of testing, I haven't seen any regressions.

rhuss:
ah, cool. So my strategy is now to enable streaming / chunking by default, add a (undocumented) option for reverting to the old behaviour in case of issues and remove the fallback code after some releases.

I will add this to 1.3.4.

Thanks again !

rhuss:
@gnufied Hi ;-) Sorry for pinging you again, but I have a question about your contribution https://github.com/rhuss/jolokia/blob/master/agent/core/src/main/java/org/jolokia/util/ChunkedWriter.java . It looks like that it takes over some parts or is derived directly from java.nio.StreamEncoder .

If you can confirm this, we have to replace this as `StreamEncoder` is licensed under the GPL which is not compatible with the license we are using (APL). 

Thanks, and sorry for bugging you again. Don't get me wrong, your contribution is very valuable but we need to get this sorted out. It would be awesome if you could reply soon.

thanks ...
... roland

gnufied:
@rhuss I can confirm that bits of `ChunkedWriter` was used from `StreamEncoder` because it implements similar interface. 

I can revisit deriving from `StreamEncoder` rather than using bits of code directly from the class. I think it should be possible. Let me look into that over the weekend.

rhuss:
@gnufied Awesome, thanks a lot !

rhuss:
@gnufied had you any change to look into this ?

agent/jvm/src/main/java/org/jolokia/jvmagent/handler/JolokiaHttpHandler.java
```diff
@@ -317,6 +318,47 @@ private void sendForbidden(HttpExchange pExchange, SecurityException securityExc
...
+
+    private void sendStreamingResponse(HttpExchange pExchange, ParsedUri pParsedUri, JSONStreamAware pJson) throws IOException {
+        ChunkedWriter writer = null;
+        try {
+            Headers headers = pExchange.getResponseHeaders();
+            if (pJson != null) {
+                headers.set("Content-Type", getMimeType(pParsedUri) + "; charset=utf-8");
+                String callback = pParsedUri.getParameter(ConfigKey.CALLBACK.getKeyValue());
+                pExchange.sendResponseHeaders(200, 0);
+                writer = new ChunkedWriter(pExchange.getResponseBody(), "UTF-8");
```
rhuss: @gnufied one question: Why actually it is required to use a `ChunkedWriter` here ? Or stated otherwise: What would be the drawback to use the response body outputstream directly when streaming out the response ?

I'm still looking for a quick solution to get rid of the IP issue we have (and which has been escalate quite a bit), so if don't find a quick solution I'm going to remove ChunkedWriter altogether this weekend.

rhuss: @gnufied I mean, replacing `ChunkedWriter` with `OutputStreamWriter`, what would we loose or what would be the issues ?

rhuss: @gnufied If I understand the code right, then the only difference between a plain `OutputStreamWriter` and `ChunkedWriter` is, then the latter does always "full" flush on `flush()` whereas the `OutputStreamWriter` does this during `close()`. 

Why is this required ?

gnufied:
@rhuss I will look into this tonight. I haven't had a chance to look yet.

rhuss:
@gnufied cool, thanks ! sorry for bugging you, but let's get that elephant out of the room ;-) Going to make a 1.4.0 release short after the fix ...

agent/jvm/src/main/java/org/jolokia/jvmagent/handler/JolokiaHttpHandler.java
gnufied: There are other minor differences - `ChunkedWriter` writes a empty byte on flush , the OutputStreamWriter doesn't - https://github.com/rhuss/jolokia/blob/master/agent/core/src/main/java/org/jolokia/util/ChunkedWriter.java#L117

gnufied: So partly this is coming back to me(sorry I forgot bunch of this code) - the main reason we had to reimplement chunkedWriter from scratch is because of empty byte handling on flush. 

StreamEncoder has `OutputStream` property marked as private  `OutputStreamWriter` delegates to `StreamEncoder` for majority of its functionality anyways.

So the problem boils down to - we need to write the empty byte on flush, to mark it as end of the whole message. `OutputStreamWriter` does not do that and hence its chunking implementation is  weirdly incomplete.

Now - in `StreamEncoder` because `OutputStream` property is private, a subclass can't access it to write that empty byte. Also directly calling `streamEncoder.write(EMPTY_BYTE)` will result in message getting buffered and that is not what we want, because this was supposed to be end of whole message. We want to be able to directly write the `EMPTY` byte to `OuputputStream`.

I think that was the most crucial reason. There may be something else but I don't remember it.

rhuss: @gnufied thanks a lot for the explanation. However, I don't get why you need to write an empty byte array to the stream. Whit is this needed ? 

Also, if I'm not completely wrong, this write is a no-op anyway:

```java
private static final byte[] EMPTY = {};
...
out.write(EMPTY);
```

then in `OutputStream`:

```java
  public void write(byte b[]) throws IOException {
        write(b, 0, b.length);
   }
   ....
 public void write(byte b[], int off, int len) throws IOException {
        if (b == null) {
            throw new NullPointerException();
        } else if ((off < 0) || (off > b.length) || (len < 0) ||
                   ((off + len) > b.length) || ((off + len) < 0)) {
            throw new IndexOutOfBoundsException();
        } else if (len == 0) {
             // This is where we end up when writing an empty array. So the original 
            // operation is a complete no-op. What did I miss ?
            return;
        }
        for (int i = 0 ; i < len ; i++) {
            write(b[off + i]);
        }
    } 
```

gnufied: I need to double check but that final EMPTY byte array causes "0\r\n\r\n" to be written to the message and that is essential for chunking to be finished.

Issue #247, opened 2016-03-09
Allow self-signed SSL certificates


gquintana:
`J4pClientBuilder` can hardly be extended to tune HTTP Client connection.
Basically, making `J4pClientBuilder#getSocketFactoryRegistry` method protected would allow to inject a custom `TrustManager` in the `SSLContext`.

PR #248, opened 2016-03-10
Make SSL connection configurable

gquintana:
See issue #247


rhuss:
Thanks ! I changed it a bit to use a builder hook, so you dont need to subclass the builder for this sort of customization. 'hope thats ok.

PR #253, opened 2016-04-13
Issue#251 No Servlet Container Info from jolokia info

luettenberg:
Proposal for Issue#251. 
Pattern changed for flexible end of string


luettenberg:
Saw it is fixed so forget...

Issue #252, opened 2016-04-13
Add Gson/JSonPath converters


svarcheg:
Hello, 

Currently we are thinking about moving from JMX-RMI to jolokia for our servers management tooling. 

The migration is pretty smooth but i run into an issue, back in time we made an extensive usage of java beans returned/passed as a parameter to our jmx operations. Another constraint is that changing those objects sources is going to be overly complex and impossible for already released versions. 

Jolokia's parameter and return values serialization do a pretty good job for response value serialization but the parameter serialization does not work. 

So i was wondering if you can suggest any solution. 

After looking into the code i m thinking about adding an optional GSon/JSonPath serialization/deserialization.
An option could be added into JsonConvertOptions (aka useGson=true/false) and converters could be updated to handle that option properly. 

JSonPath is not a must have, but might be an optional replacement to jolokia's paths. 

Let me know what do you think. If you re ok i'll do a pull request. 

Regards, 

Vladimir

rhuss:
Thanks for your question and for your interest. 

You are right, parameter serialization is limited (as explained [here](https://github.com/rhuss/jolokia/issues/212#issuecomment-136359434), too). However, Jolokia is not meant to be a full remoting framework and one design goals for the Jolokia agents is a small footprint (still below 500k). This also is possible because beside `json-simple` (20k) there are no dependencies. jsonpath alone has 200k (+ quite some transitive dependencies), gson also 230k (+ transitive dependencies), so that this would roughly double the size of the agent (or add 50% when only adding gson). So for Jolokia 1 I don't plan changes here.

But there is hope ;-) For Jolokia 2.0 it is planned that the serializer can easily be exchanged. Jolokia 2.0 is targeted for mid-year. It is already functional and contains all features of Jolokia 1.3.3. So maybe you want to give it a try ? In order to plugin in your own serializer all you need is to implement three methods of the [Serializer](https://github.com/rhuss/jolokia/blob/2.0/server/core/src/main/java/org/jolokia/server/core/service/serializer/Serializer.java) interfaces, whose single implementation is currently [JolokiaSerializer](https://github.com/rhuss/jolokia/blob/2.0/service/serializer/src/main/java/org/jolokia/service/serializer/JolokiaSerializer.java), the serializer you know from Jolokia.

If you want to give it a try with a GSon serializer implementation I would happily integrate it for Jolokia 2.0.

svarcheg:
Thanks for the fast answer. 

[...]

PR #255, opened 2016-04-26
2.0 serializers

svarcheg:
Slightly changed Serializer interface in order to return an object containing the updated outer object and the old inner value. This is to avoid an implicit parameter change and make writing new Serializers implementations easier. 

Also removed the exception thrown when more than one mandatory service is found. Made it behave the same way as optional services (except throwing an exception when the mandatory service is missing).


svarcheg:
Hum not sure why CI failed .. does not feel like it linked the the changes ..

svarcheg:
fixed 2.0 test failures (caused by too many open ports). Please have a look when you have time.

rhuss:
Thanks a lot ! 'hope I will have time to review it over the weekend.

svarcheg:
Thanks, I look forward to hearing from you.

rhuss:
LGTM.

You enhanced the CleanupThread, too. Looks better now, but wonder whether you had any issues with that ?

rhuss:
Just curious, any news on the new serializer you had in mind ?

svarcheg:
Yup I've changed a bit the cleanup thread indeed.  The problem is that in some cases the server was not stopped and it was "leaking" open ports which made existing tests fail randomly ("Cannot find a single free port"). 

For the new serializer, it s in progress, it appears to be slightly more complex to implement as there's some implicit expectations from the Serializer (i may try to change the interface to make those more explicit). Currently working on making existing integration tests pass with new serializer.

svarcheg:
Hello, 

Added another small change in order to be able to handle type erasure in case of generic mbeans. 

Also I have a working copy of a GSon serializer, will do a pull request once this one is merged as it s based on this one. 

Please let me know if you have any comments or any other things which would prevent you from merging. 

Regards, 

Vladimir

rhuss:
Thanks, I will have a look soon.

svarcheg:
Did you have some time to look over it ? (aka gentle reminder :))

rhuss:
Thanks for the heads up ;-) I'm just back from a week of conferences and currently catching up. But its on my task list for this week.

Thanks again ...
... roland

svarcheg:
no worries, thanks

svarcheg:
Hello, 
did you have some time to look over ? 

Regards, 

Vladimir

rhuss:
Oops, confused this PR with #265 where I had [compilation errors](https://github.com/rhuss/jolokia/pull/265#issuecomment-225151331)

Is #265 still current ? If not, could you please close it.

I will have a look to **this** PR now this week.

sorry ...

rhuss:
Ok, now got it. I need both this and #265.

rhuss:
Ok, will do my very best to sort things out ;-)

svarcheg:
Yup sorry for that, #265 is based on this one, not sure there's an easy way to base a PR on another PR .. 

Also I have another commit where I completely replaced JolokiaSerializer with GSon one. It is very impacting (i have to change some tests for those to pass) so I didnt push it on github yet. I can do that at some point if you're interested in it.

rhuss:
thanks, no problem. This week I'll have the time todo the integration. thanks for your patience, and so sorry for the delay.

svarcheg:
no problems, thanks for you time

service/jmx/src/main/java/org/jolokia/service/jmx/handler/ExecHandler.java
```diff
@@ -73,12 +73,20 @@ public Object doHandleSingleServerRequest(MBeanServerConnection server, JolokiaE
...
-        		params[i] = context.getMandatoryService(Serializer.class).deserialize(types.paramClasses[i], args.get(i));
-        	}
+        List<String> clientTypes = splitOperation(request.getOperation());
+        for (int i = 0; i < nrParams; i++) {
+            if (types.paramOpenTypes != null && types.paramOpenTypes[i] != null) {
+                params[i] = context.getMandatoryService(Serializer.class).deserializeOpenType(types.paramOpenTypes[i], args.get(i));
+            } else {
+                String type;
+                String clientType = clientTypes.size() > i + 1 ? clientTypes.get(i + 1) : null;
+                if (clientType != null && !clientType.equals(types.paramClasses[i])) {
```
rhuss: In which case can the client provided type different than the parameter type obtained from JMX by doing a lookup with this client type ? Do you have an example for this ?

If this is required I would suggest to move `clientTypes` into `types` so that we can avoid the second calls to `splitOperation()`. imo this makes also sense since types is about all determined type information.

service/jmx/src/main/java/org/jolokia/service/jmx/handler/ExecHandler.java
```diff
@@ -187,9 +195,10 @@ private OperationAndParamType extractOperationTypes(MBeanServerConnection pServe
                 // Number of arguments dont match
                 continue OUTER;
             }
-            for (int i=0;i<infos.length;i++) {
+            for (int i = 0; i < infos.length; i++) {
                 String type = infos[i].getType();
-                if (!type.equals(pTypes.get(i))) {
+                // Object is here for an eventual case of type erasure
+                if (!type.equals(pTypes.get(i)) && !type.equals(Object.class.getName())) {
```
rhuss: Do you have an example where type erasure cause `MBeanParameterInfo.getType()` to return `Object` ?

service/jmx/src/main/java/org/jolokia/service/jmx/handler/WriteHandler.java
```diff
@@ -119,7 +120,7 @@ private Object setAttribute(JolokiaWriteRequest request, MBeanServerConnection s
      * @throws IllegalAccessException if access to MBean fails
      * @throws InvocationTargetException reflection error when setting an object's attribute
      */
-    private Object[] getValues(String pType, Object pCurrentValue, JolokiaWriteRequest pRequest)
+    private WriteRequestValues getValues(String pType, Object pCurrentValue, JolokiaWriteRequest pRequest)
```
rhuss: I see you point to avoid an untyped Object[] array as return value. However, I wonder whether we could make `WriteRequestValue` a inner class here and avoid it in the `Serializer` interface.

service/jmx/src/main/java/org/jolokia/service/jmx/handler/WriteHandler.java
```diff
@@ -132,21 +133,15 @@ private Object setAttribute(JolokiaWriteRequest request, MBeanServerConnection s
 
             // We set an inner value, hence we have to return provided value itself for resetting
             // it later back via JMX
-            return new Object[] {
-                    pCurrentValue,
-                    context.getMandatoryService(Serializer.class).setInnerValue(pCurrentValue, newValue, pathParts)
-            };
+            return context.getMandatoryService(Serializer.class).setInnerValue(pCurrentValue, newValue, pathParts);
```
rhuss: We could use here also a `new WriteRequestValues()` like in the else branch, so that `setInnerValue()` can still return the old value only. The reason why I don't like to put WriteRequestValues into the interface is that is (a) a new object, and (b) contains the original value `newValue` given as argument. People could be confused which to use later on because it is no **guaranteed** that both are the same. If you only have one (the method argument), this is unambiguous.

agent/jvm/src/main/java/org/jolokia/jvmagent/CleanupThread.java
```diff
@@ -68,9 +68,10 @@ public void run() {
     /**
      * Stop the server.
      */
-    public void stopServer() {
+    public void stopServer() throws InterruptedException {
         active = false;
         interrupt();
+        join(); // once interrupted, wait for server to stop
```
rhuss: Thanks, makes sense for 1.x, too ;-)

I will backport this to Jolokia 1.x if you dont mind.

rhuss:
Looks good to me, thanks a lot.

 Not yet sold on the API change, though, but I guess I have to look in you GSon implementation first. Which I will do now ;-)

service/jmx/src/main/java/org/jolokia/service/jmx/handler/ExecHandler.java
svarcheg: I added a test case below instead of example, i guess it s better. Let me know if something is not clear in there. 

About you suggestion, yup it looks better than my initial proposal will push another change for that.

service/jmx/src/main/java/org/jolokia/service/jmx/handler/ExecHandler.java
svarcheg: Same as above please check the test I added below, and do not hesitate to tell me if something is unclear.

agent/jvm/src/main/java/org/jolokia/jvmagent/CleanupThread.java
svarcheg: yup no problem

service/jmx/src/main/java/org/jolokia/service/jmx/handler/WriteHandler.java
svarcheg: Hum, i m not especially attached to typing things here. I ll try to explain the problem I attempted to solve so maybe you ll have better ideas. 

When implementing the gson/jsonpath based serializer I was confused about how the setInnerValue was working since only the old value was returned. It took me some time to figure out that the return value does not have a lot of importance and the method in `JolokiaSerializer` is basically mutating the `pOuterObject` and then the caller sets it properly on a bean. 

This approach is valid but still it makes quite complex (and long) to implement a custom serializer without reimplementing all the reflexion mechanics JolokiaSerializer has to find out the right operation and so on. 

This is why I tried to give setInnerValue a bit more function-like behavior taking some read-only parameters and returning a value (a pair of values actually). This approach makes things easier for the serializer implementor.

If you do prefer keeping the "untyped" approach we could easily replace WriteRequestValues by an `Object[]`. 

Please let me know what do you think.

service/jmx/src/main/java/org/jolokia/service/jmx/handler/WriteHandler.java
svarcheg: I m not really attached to WriteRequestValue class, we can stick to Object[] if you prefer, my only blocking point is described below :)

svarcheg:
Rebased the PR on a current 2.0 and added a test case example. Please not hesitate to tell me what do you think. 

Regards, 

Vladimir

rhuss:
Thanks ! a bit busy right now, hope to get to review your update this week.

svarcheg:
Hello, 

Did you have time to have a look ? If you have any questions, will be happy to answer.

Regards, 

Vladimir

rhuss:
sorry, work is eating me ;-) but I will have a look soon (as I have to finalize 2.0 soon, too).

thanks for your patience ...

svarcheg:
yup no worries ;)

agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServer.java
```diff
@@ -149,7 +149,11 @@ public void stop() {
         serviceManager.stop();
 
         if (cleaner != null) {
-            cleaner.stopServer();
+            try {
+                cleaner.stopServer();
+            } catch (InterruptedException exp) {
+                System.err.println("Could not stop Jolokia server properly: " + exp);
```
rhuss: I wouldnt print to stderr here, since JolokiaServer is also used by other directly, so its not sure in which circumstance its running. Instead the `LogHandler` should be used because it can be muted, too. (Add a member variable `log` in `init()` for this)

service/jmx/src/main/java/org/jolokia/service/jmx/handler/ExecHandler.java
rhuss: I see your point, but this is still not necessary here, since `extractOperationTypes()` does this already. E.g. when I remove this branch, the unit tests still succeeds.

rhuss: Sorry my bad. You are right, we should keep that

All good ;-)

service/jmx/src/main/java/org/jolokia/service/jmx/handler/WriteHandler.java
rhuss: Sorry, I still didn't get it ;-) The first argument to `setInnerValue` is the object on which to set a property, either a direct property or one which is nested deeper when a path is used. This object, which you provide as argument, is returned _unchanged_ as part of the return value (in `WriteRequestValue`). 

At least this is the case for the [default serializer](https://github.com/rhuss/jolokia/blob/2.0/service/serializer/src/main/java/org/jolokia/service/serializer/JolokiaSerializer.java#L77-L79). If you can convince me that:

* there are situations where `WriteRequestValue.oldValue` is different from `pCurentValue` (the first argument to `setInnerValue()`
* how it helps in diving into the object when a path is given (e.g. to avoid reflection)

then I'm all for it. Otherwise its only boilerplate, if the value given as argument is always returned because a called can easily remember the argument to work with it after the call to the serializer.

 But I guess I'm just a bit lamebrained ;-)

rhuss:
Just had a review and it almost LGTM ;-). Thanks a lot for the work (and the patience with me ;-)

There a two minor things we need to resolve:

* The `System.out.error` in JolokiaServer (not a biggy I guess)
* The `setInnerValue()` thingy. I've the feeling, I really missed sth. If you can convince me with an example e.g. from your serializer how it makes your live easier, I'm all for it. And then I would also add it as a typed object as you did. 

Thanks again, I promise the next turaround goes faster ;-). I already have some idea how we can package up a second serializer (or any other alternative component for Jolokia), more on this later.

svarcheg:
Hello, 

Sorry for such a late answer I was on summer vacations :) 

I'll fix that System.error thing (shouldn't be too hard indeed). 

As for setInnerValue as you stated in the #252 comment it's not directly related to serialization and the best solution would be to remove it from the interface (it would also solve that WriteRequestValue thingy :)). This being said I doubt that it is possible in a reasonable amount of time. So for now I would go for allowing  UnsupportedOperation being thrown from setInnerValue and will remove WriteRequestValue. 

Will try to update the PR this week. 

Kind regards, 

Vladimir

svarcheg:
Hello, 

Did you have some time to look over it ?

Regards,

rhuss:
Sorry, not yet. But will do it. In the meantime, could you please have a look at  https://github.com/rhuss/jolokia/pull/265#r73579774 and maybe comment on it ? That's still the question whether we really should add `setInnerValue()` as its really only for a special uses case which maybe doesnt happen that often but needs a high implementation effort.

svarcheg:
Hello

bump :) do you think you can review this at some point ? 

Kind regards, 

Vladimir

tadayosi:

PR #257, opened 2016-05-13
Allow user-specifiable SSL / TLS protocols and ciphers to the JVM agent

drcapulet:


rhuss:
Thanks a lot !

Sorry for the delay, quite busy these days. But it looks like a valuable addition which make sense. I will be off next week, but hopefully can dive into it the week after.

rhuss:
Looks good to me. Thanks also for the doc update and fixes.

And so sorry for the long delay, there where busy times ;-(. There will be a 1.3.4 release at the end of the week or beginning of next week with these changes applied.

Thanks again ....

drcapulet:
No worries, thanks @rhuss!

Issue #258, opened 2016-05-17
How to implement compatibility of the Jolokia JVM agent with org.jboss.logmanager.LogManager loaded form JBoss modules


rraez:
## Problem

Currently we have enabled the Jolokia JVM agent with JBoss EAP 6.0.x & 6.2.x.

To get the JVM Agent running, we have added the `jboss-logmanager.jar` the the `-Xbootclasspath`. 

This is currently the official workaround described by RedHat: https://access.redhat.com/solutions/312453

Alos other JVM agents have the same problem: https://issues.jboss.org/browse/WFLY-895

But in JBoss EAP 6.4. this is no more working:

```
java.lang.IllegalStateException: The LogManager was not properly installed (you must set the "java.util.logging.manager" system property to "org.jboss.logmanager.LogManager")
    at org.jboss.logmanager.Logger.getLogger(Logger.java:58)
    at org.jboss.as.server.Main.main(Main.java:84)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.jboss.modules.Module.run(Module.java:312)
    at org.jboss.modules.Main.main(Main.java:473)
```

Even when you add the system property `java.util.logging.manager=org.jboss.logmanager.LogManager` and add the `jboss-logmanager.jar` the the -Xbootclasspath, the following code throws a class cast exception:
https://github.com/jboss-logging/jboss-logmanager/blob/master/src/main/java/org/jboss/logmanager/Logger.java#L57

This exception is caused by changed class loading behaviour in JBoss eap 6.4. It looks like the `org.jboss.logmanager.Logger` class is loaded a second time from the modules and therefore it can't be casted.

In my option it was never a clean solution to override classes loaded by JBoss modules via the system class loader (`-Xbootclasspath`). 

I have to find a solution for already released version of JBoss eap (6.2.2, 6.4.1, 6.4.7).
## Implement a solution for the problem in products is not easy

I agree with you suggestion that a solution within the product (e.g. [swarm](https://issues.jboss.org/browse/SWARM-204) ) would be nice, but it is not that simple, as the `java.util.logging.LogManager` of java is a singleton initialized the first time it is used.
[...]

PR #260, opened 2016-05-20
Compatibility of the Jolokia JVM agent with org.jboss.logmanager.LogManager loaded form JBoss modules #258

rraez:
- Extended the ServerDetector interface for the early detection of a server.
- The JBossDetector now detects JBoss modules based containers and waits until the custom java LogManager is initialized.
- The JvmAgent is now creating another daemon to launch the startup of in the background. 

No addressed:
- Currently if the early detection in the JBossDetector kicks off, there will be an endless loop. A timeout is not a good option. What do you think about to add feature to JvmAgentConfig to be able to skip earlyDetect & awaitServerInitialization for a detector?
- There is no integration wit the CleanupThread managed in the JolokiaServer. This would require some refactoring. I'm not sure if this is needed.
- I have only performed some integration tests with JBoss EAP, I have not tested any other containers.


rhuss:
Thanks a lot ! 'will review it soon, hopefully this week.

rhuss:
Sorry, still late ;-( Unfortunately I'm off next week for some conferences, but I try to catch up this the week after.

agent/jvm/src/main/java/org/jolokia/jvmagent/JvmAgent.java
```diff
@@ -73,29 +78,60 @@ public static void premain(String agentArgs) {
...
+     * @param instrumentation 
+     * @see ServerDetector#earlyDetect(Instrumentation)
+     * @see ServerDetector#awaitServerInitialization(Instrumentation)
+     */
+    private static void awaitServerInitialization(JvmAgentConfig pConfig, final Instrumentation instrumentation) {
+        List<ServerDetector> detectors = MBeanServerHandler.lookupDetectors();
+        for (ServerDetector detector:detectors) {
+            // TODO add feature to JvmAgentConfig to be able to skip earlyDetect & awaitServerInitialization 
+            // for a detector. 
+            if (detector.earlyDetect(instrumentation)) {
```
rhuss: I wonder whether we could combine `earlyDetect` and `awaitServerInitialization` into a single method e.g. `serverPreInit`. The JBossServer detector could do the detection and waiting within this method, so we could remove one method from `ServerDetector`.

On the other end, we could use `earlyDetect()` to start only a thread if a detector returns true. I'm not sure whether there could be side effects when the Jolokia JVM HttpServer and the normal initialization continue in parallel (or when an external requests comes in early). But I guess that should be no problem, so I tend to reduce this to a single method.

agent/core/src/main/java/org/jolokia/detector/JBossDetector.java
```diff
@@ -33,6 +35,59 @@ public ServerHandle detect(MBeanServerExecutor pMBeanServerExecutor) {
...
+                    // is or will be initialized.
+                    // Here where trigger to for load the class:
+                    // https://github.com/jboss-modules/jboss-modules/blob/1.5.1.Final/src/main/java/org/jboss/modules/Main.java#L482
+                    // Therefore the steps 3-6 of the proposal for option 2 don't need to be performed,
+                    // see https://github.com/rhuss/jolokia/issues/258 for details.
+                    break;
+                }
+            }
+            try {
+                Thread.sleep(200);
```
rhuss: In fact I would add nevertheless some sort of (long) default timeout as a last resort and log an error. Maybe sth like 5 minutes ?

rhuss:
LGTM, thanks ! (and sorry for the quite long delay, was a busy time ;-)

> Currently if the early detection in the JBossDetector kicks off, there will be an endless loop. A timeout is not a good option. What do you think about to add feature to JvmAgentConfig to be able to skip earlyDetect & awaitServerInitialization for a detector?

Afais only the JBossDetector would be affected so for other servers there should be no problem at all. In fact, I would nevertheless add a long timeout to at least bailout after some time. Could be as long as 10 minutes. I think initialization of the logging system should happen always quite fast, independent of the applicatins to start. Maybe one could also log out how long the wait was actually ?

> There is no integration wit the CleanupThread managed in the JolokiaServer. This would require some refactoring. I'm not sure if this is needed.

Don't think its needed since the thread starting the server will only start another thread and then ends [here](https://github.com/rhuss/jolokia/blob/master/agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServer.java#L128). So this startup thread should be done quite early so that no cleanup should be required. And as it is marked as a daemon thread I dont expect much trouble.

> I have only performed some integration tests with JBoss EAP, I have not tested any other containers.

I can do later some tests on older JBoss versions. Would be interesting, too, how it works with Wildfly Swarm as Wildfly Swarm has the same [issue](https://issues.jboss.org/browse/SWARM-204).

PR #263, opened 2016-06-01
[WIP] Hot Threads-POC based on that of Elasticsearch

alexbrasetvik:
Hi,

I realise this probably isn't mergable as is, so consider this PR a request for feedback.

Elasticsearch has a [hot_threads-API](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-hot-threads.html) that is very useful to hit to debug what a misbehaving JVM is up to. It would be awesome to be able to add that via Jolokia to an arbitrary process.

Making it a custom action like this proof-of-concept is probably not the right way to go. Would it make more sense to make it an MBean? If so, is that something you could consider accepting into core Jolokia, or should it be made a plugin?

Thanks for the great work with Jolokia :)

### DIFF
--- a/agent/core/src/main/java/org/jolokia/config/ConfigKey.java
+++ b/agent/core/src/main/java/org/jolokia/config/ConfigKey.java
@@ -307,7 +307,10 @@ public enum ConfigKey {
      * this can be used by clients to provide additional information to
      * the user.
      */
-    AGENT_DESCRIPTION("agentDescription",true,false);
+    AGENT_DESCRIPTION("agentDescription",true,false),
+
+    HOT_THREADS_INTERVAL("thread_duration", false, true, "500"),
+    HOT_THREADS_COUNT("number_of_threads", false, true, "10");
 
     /**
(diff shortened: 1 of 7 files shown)


rhuss:
In fact Jolokia is supposed to be a transparent JMX bridge without any knowledge of the underlying metrics. You can see this also in the protocol which doesn't refer to any specific metrics but are realized in very general terms.

Your second suggestion makes much more sense. Adding this extra functionality as an MBean which in turn could expose a `JsonObject` is the way to go. That way you can call a JMX operation and get back a rich JSON response.

The question then is where and when to register this MBean. This is probably best done when starting up Elasticsearch, but there can be other options, too. BTW, there's a [jolokia-extra](https://github.com/rhuss/jolokia-extra) project for adding this sort of special functionality.

I suggest that your refactor this as an MBean and then we could decide where to put it into jolokia-extra (if you want to)

rhuss:
I you don't mind I'd like to close this PR for the reasons mentioned above. Feel free to reopen it for further discussions.

PR #269, opened 2016-06-25
issue-268 - fix build errors when building via cmd line

robstryker:


robstryker:
I've just tried again by re-checking out master and performing 
mvn  -Dmaven.repo.local=/tmp/repo clean install

This seems to work. But when I remove the -Dmaven.repo.local sysprop, it fails again. 

This patch appears to fix it, though I'm not entirely sure why. 

[rob@rawbdor jolokia] (master)$ mvn -v
Apache Maven 3.2.5 (NON-CANONICAL_2015-04-01T06:42:27_mockbuild; 2015-04-01T02:42:27-04:00)
Maven home: /usr/share/maven
Java version: 1.8.0_05, vendor: Oracle Corporation
Java home: /home/rob/apps/java/java8/jdk1.8.0_05/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.4.13-200.fc22.x86_64", arch: "amd64", family: "unix"

agent/jvm/pom.xml
```diff
@@ -68,6 +75,13 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
```
rhuss: I really wonder why this dependency would be necessary as the JVM agent doesnt use servlets at all (but the JVM internal HTTP server with its own custom API)

rhuss:
I would prefer to fix the original issues you have. The build runs on Travis CI for every branch without issues, as well as it runs locally. 

I really believe is that you have a broken local ~/.m2 repository. I'll close this PR therefor, please feel free to reopen it, if you still feel Jolokia's build has still an issue.

PR #272, opened 2016-07-05
If a client cert matches one of the allowed principles, then allow th…

chirino:
…e auth to succeed


agent/jvm/src/main/java/org/jolokia/jvmagent/handler/JolokiaHttpsHandler.java
```diff
@@ -112,13 +112,16 @@ private void checkCertForAllowedPrincipals(HttpsExchange pHttpsExchange) {
             try {
                 certPrincipal = (X500Principal) pHttpsExchange.getSSLSession().getPeerPrincipal();
                 Set<Rdn> certPrincipalRdns = getPrincipalRdns(certPrincipal);
+                boolean matchFound = false;
                 for (LdapName principal : allowedPrincipals) {
-                    for (Rdn rdn : principal.getRdns()) {
-                        if (!certPrincipalRdns.contains(rdn)) {
-                            throw new SecurityException("Principal " + certPrincipal + " not allowed");
-                        }
+                    if( certPrincipalRdns.containsAll(principal.getRdns()) ) {
```
rhuss: cool, didn't remembered containsAll() (shame on me). Much better ....

PR #273, opened 2016-07-05
Moved client cert checking into Authenticator

chirino:
Also support enabling client cert and other auth modes at the same time.

### DIFF
--- a/agent/jvm/src/main/java/org/jolokia/jvmagent/security/AllAuthenticator.java
+++ b/agent/jvm/src/main/java/org/jolokia/jvmagent/security/AllAuthenticator.java
@@ -1,0 +1,61 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jolokia.jvmagent.security;
+
+import com.sun.net.httpserver.Authenticator;
+import com.sun.net.httpserver.HttpExchange;
+
+import java.util.ArrayList;
+
+/**
+ * Authenticator that succeeds if all of the Authenticators it delegates to succeed.
+ *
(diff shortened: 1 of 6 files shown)


agent/jvm/src/main/java/org/jolokia/jvmagent/security/ClientCertAuthenticator.java
```diff
@@ -112,16 +116,13 @@ private void checkCertForAllowedPrincipals(HttpsExchange pHttpsExchange) {
...
                 Set<Rdn> certPrincipalRdns = getPrincipalRdns(certPrincipal);
-                boolean matchFound = false;
                 for (LdapName principal : allowedPrincipals) {
-                    if( certPrincipalRdns.containsAll(principal.getRdns()) ) {
-                        matchFound = true;
-                        break;
+                    for (Rdn rdn : principal.getRdns()) {
+                        if (!certPrincipalRdns.contains(rdn)) {
+                            throw new SecurityException("Principal " + certPrincipal + " not allowed");
+                        }
```
rhuss: I will take over the fix from the other PR here.

agent/jvm/src/main/java/org/jolokia/jvmagent/security/AllAuthenticator.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.ArrayList;
+
+/**
+ * Authenticator that succeeds if all of the Authenticators it delegates to succeed.
+ *
+ * @author roland
+ * @since 26.05.14
+ */
+public class AllAuthenticator extends Authenticator {
```
rhuss: Just wondering whether an AllAuthenticator makes much sense. Also, it is not used currently, so maybe for KISS reasons we simply omit it now and can easily reintroduce it we allow for a more flexible configuration where an AllAuthenticator could be selected.

rhuss:
LGTM, thanks ;)

I will update the documentation accordingly. A new Jolokia release (1.3.4) is expected at the end of this or beginning of next week.

rhuss:
There will be still a problem for our use case, since the client certificate will be validated still that is signed by a CA. This happens before the authenticators are called, when setting up the [HTTPS subsystem](https://github.com/rhuss/jolokia/blob/7a8249cc1e36680aee26a8ddd570520eef3e190b/agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServer.java#L415-L415). 

So I'm afraid that its not possible to have SSL with allowing both, client cert authentication and basic authentication without a prior client cert verification.

The authenticator is used only for some extra checks, like whether the presented certificate is indeed a client certificate (and not a server certificate), and whether the enclosed principal matches a configured value.

The check, whether a client cert is presented and whether it is signed by a given CA (stored in the keystore) happens before, internally in the HttpsServer which we cant influence much except for configuring it.

chirino:
Yeah I think that limitation is ok.  In the case a client does not have cert trusted by the CA, he should not do SSL with the cert at all.

rhuss:
But how can you provide a Pod specific client cert signed by the OpenShift CA when running from a Pod ?

PR #275, opened 2016-07-15
Follow up to PR #21, lets use setWantClientAuth() instead of setNeedC…

chirino:
…lientAuth() so that you can do basic auth over SSL.  The auth will fail if client cert auth is enabled, but a valid client cert is not found.

### DIFF
--- a/agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServer.java
+++ b/agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServer.java
@@ -412,8 +412,8 @@ public class JolokiaServer {
             // get the default parameters
             SSLParameters defaultSSLParameters = context.getDefaultSSLParameters();
 
-            params.setNeedClientAuth(serverConfig.useSslClientAuthentication());
-            defaultSSLParameters.setNeedClientAuth(serverConfig.useSslClientAuthentication());
+            params.setWantClientAuth(serverConfig.useSslClientAuthentication());
+            defaultSSLParameters.setWantClientAuth(serverConfig.useSslClientAuthentication());
 
             // Cipher Suites
             params.setCipherSuites(serverConfig.getSSLCipherSuites());


agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServer.java
```diff
@@ -412,8 +412,8 @@ public void configure(HttpsParameters params) {
             // get the default parameters
             SSLParameters defaultSSLParameters = context.getDefaultSSLParameters();
 
-            params.setNeedClientAuth(serverConfig.useSslClientAuthentication());
-            defaultSSLParameters.setNeedClientAuth(serverConfig.useSslClientAuthentication());
+            params.setWantClientAuth(serverConfig.useSslClientAuthentication());
+            defaultSSLParameters.setWantClientAuth(serverConfig.useSslClientAuthentication());
```
rhuss: Without adapting the Authenticators later on this would allow people to pass authentication when 

* `useSslClientAuthentication` is true
* No `clientPrincipal` given
* No `extendedClientCheck` set
* No client cert presented.

This shouldn't be the case. Looking into the `ClientCertAuthenticator` its even worse, since the `extendedClientCheck` and `clientPrincipal` verification [is only triggered when a client cert is presented](https://github.com/rhuss/jolokia/blob/69586e7d6321748a27528013b435763bd45d5c03/agent/jvm/src/main/java/org/jolokia/jvmagent/security/ClientCertAuthenticator.java#L87-L88).

But I think you are on the right track here, we just need to verify in `ClientCertAuthenticator` that a valid client certificate, signed by the configured CA (but this hopefully already verified by the SSLEngine then), is presented.

rhuss:
From http://stackoverflow.com/questions/14875094/ssl-server-socket-want-auth-option its unfortunately not really clear to me, whether a client can present an untrusted client cert (i.e. not signed by the CA) and still pass with `wantClientAuth`. It seems that it gets rejected by the SSL engine, but if not, we need to verify any presented cert with the CA's public key on our own.

rhuss:
- Added this to master and adapted the ClientCertAuthenticator that it returns a Failure when not client cert is presented at all
- Added some unit testst
- Updated changes.

I'm going to prepare a next 2.0-M4 milestone and hopefully release a 1.3.4 this week with these changs

rhuss:
@chirino : Jolokia 2.0.0-M3 is released, the s2i images has been upgraded an a fabric8/s2i-java-jboss:1.2.11 has been pushed to index.docker.io

rhuss:
@chirino : Please let me know, whether this works for you so that I can eventually do a Jolokia 1.3.4 release.

rhuss:
@chirino Could you test it ?

rhuss:
@chirino ok, I suppose its ok for you. Going to release now 1.3.4. overdue anyway.

Issue #276, opened 2016-07-19
Remote policyLocation with Basic-Auth


mweirauch:
Hi,

does it look feasible to add support for Basic-Auth configuration when the `policyLocation` is retrieved from a remote location?

Background: We are using `spring-boot` with centralized configuration management via `spring-cloud-config`. The latter allows for serving "raw" files via it's interface. Think `jolokia-access.xml` served from the config server.

`spring-boot` allows setting the servlet init parameters. Thus the following setup:

```
jolokia.config.policyLocation=http://configserver/*/*/*/jolokia-access.xml
```

Unfortunately our config server is guarded with HTTP Basic-Auth and so the `RestrictorFactory` will get a `401` response and fail. Also the Java `UrlConnection` doesn't like URL tricks like `http://user:pass@host` so this isn't an option either. One would have to set the Basic-Auth headers itself.

A setup of the config server without Basic-Auth and jolokias Restrictor configured from remote works like a charme.

The config-server way would be my ideal solution instead of distributing a jolokia-access.xml with every application. (Be it by dependency, a spring boot starter or a build step for making sure the file is within the deployable artefact.)

rhuss:
I could image to provide a functionality like this by parsing the URL for user & password and the setting the `Authorization:` header accordingly on our own. This would save some new configuration keys from which it already have too much (so that I always hesitate a bit to introduce new keys). 

However, I wonder whether this really make sense when used together with plain `http://`. After all `jolokia-access.xml` is about security and if you send it over http this can be easily forged with a man-in-the-middle attack. And since the password is send via plain text as well, this could be grabbed, too.

IMO it only makes sense to either use SSL (and then with server cert validation), or to put the restrictions really into the jar where it does not need to be loaded over the net during runtime.

BTW, you can easily write your own restrictor, too, which can do what you want. It depends on the agent your are using how to fire it up. You simply specify a classname with `restrictorClass` which needs to be in the classpath. You need to implement [Restrictor](https://github.com/rhuss/jolokia/blob/master/agent/core/src/main/java/org/jolokia/restrictor/Restrictor.java) and provide a constructor which takes a [Configuration](https://github.com/rhuss/jolokia/blob/master/agent/core/src/main/java/org/jolokia/config/Configuration.java) as single argument. 

[...]

PR #279, opened 2016-08-05
#276 : Basic-Auth support for policyLocation

mweirauch:
This took a little longer due to other work related tasks...

Ideally I'd have wanted to build a separate "UrlBuilder" unit but that would have also meant providing it to another package private factory method for unit testing it's orchestration. Thought it was too much overhead and skipped that.
- A `rawLocation` had to be introduced in order to not print out the env-interpolated URL.
- Also testing URLConnection for presence of the header field is tricky as the `Authorization` header is regared secure by the Sun implementation and thus not returned on query. That's why the URL building return object which is primarily used for the tests.

I've likely broken a dozen style guidelines as well. Not sure I've been creative enough regarding the unit test URLs.

### DIFF
--- a/agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
+++ b/agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
@@ -4,7 +4,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.net.MalformedURLException;
 import java.net.URL;
+import java.net.URLConnection;
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.jolokia.config.ConfigKey;
 import org.jolokia.config.Configuration;
@@ -35,6 +40,27 @@ import org.jolokia.util.*;
  */
 public final class RestrictorFactory {
 
+    private static final Pattern URL_PATTERN = Pattern.compile("(http[s]?://)((.+):(.+)@)?(.+)");
+
+    /* default */ static class URLConnectionResult {
+
+        public final URLConnection urlConnection;
+
+        public final String url;
+
+        // URLConnection isn't queryable for "security headers";
+        // make this available for unit testing
+        public final String basicAuthHeaderValue;
+
+        private URLConnectionResult(URLConnection urlConnection, String url,
+                String basicAuthHeaderValue) {
+            this.urlConnection = Objects.requireNonNull(urlConnection);
+            this.url = Objects.requireNonNull(url);
+            this.basicAuthHeaderValue = basicAuthHeaderValue;
+        }
+
+    }
+
     private RestrictorFactory() { }
 
     public static Restrictor createRestrictor(Configuration pConfig, LogHandler logHandler) {
@@ -45,18 +71,19 @@ public final class RestrictorFactory {
             return customRestrictor;
         }
 
-        String location = NetworkUtil.replaceExpression(pConfig.get(ConfigKey.POLICY_LOCATION));
+        String rawLocation = pConfig.get(ConfigKey.POLICY_LOCATION);
+        String location = NetworkUtil.replaceExpression(rawLocation);
         try {
             Restrictor ret = RestrictorFactory.lookupPolicyRestrictor(location);
             if (ret != null) {
-                logHandler.info("Using policy access restrictor " + location);
+                logHandler.info("Using policy access restrictor " + rawLocation);
                 return ret;
             } else {
                 logHandler.info("No access restrictor found, access to any MBean is allowed");
                 return new AllowAllRestrictor();
             }
         } catch (IOException e) {
-            logHandler.error("Error while accessing access restrictor at " + location +
+            logHandler.error("Error while accessing access restrictor at " + rawLocation +
                              ". Denying all access to MBeans for security reasons. Exception: " + e, e);
             return new DenyAllRestrictor();
         }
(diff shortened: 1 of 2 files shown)


agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
```diff
@@ -35,6 +40,27 @@
...
+
+        public final String url;
+
+        // URLConnection isn't queryable for "security headers";
+        // make this available for unit testing
+        public final String basicAuthHeaderValue;
+
+        private URLConnectionResult(URLConnection urlConnection, String url,
+                String basicAuthHeaderValue) {
+            this.urlConnection = Objects.requireNonNull(urlConnection);
```
rhuss: Sorry, we can't use `Objects` since its only available since Jdk 1.7. Jolokia is supposed to support still 1.6 (yes I know, but there are still many 1.6 users outthere). For Jolokia 2.0 we will relax this then to at least Java 7 (or maybe alread 8)

agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
```diff
@@ -45,18 +71,19 @@ public static Restrictor createRestrictor(Configuration pConfig, LogHandler logH
...
         }
 
-        String location = NetworkUtil.replaceExpression(pConfig.get(ConfigKey.POLICY_LOCATION));
+        String rawLocation = pConfig.get(ConfigKey.POLICY_LOCATION);
+        String location = NetworkUtil.replaceExpression(rawLocation);
         try {
             Restrictor ret = RestrictorFactory.lookupPolicyRestrictor(location);
             if (ret != null) {
-                logHandler.info("Using policy access restrictor " + location);
+                logHandler.info("Using policy access restrictor " + rawLocation);
```
rhuss: Any particular reason why you dont want to log the resolved location, but the specified one ?

agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
```diff
@@ -35,6 +40,27 @@
  */
 public final class RestrictorFactory {
 
+    private static final Pattern URL_PATTERN = Pattern.compile("(http[s]?://)((.+):(.+)@)?(.+)");
+
+    /* default */ static class URLConnectionResult {
+
+        public final URLConnection urlConnection;
```
rhuss: can be package visisble (since tests are in the same package, too)

agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
mweirauch: Yikes. Made the same mistake with Base64 and switched to Base64Util virtually just before commit. Whyever maven is not nagging as source and target source level are set to 1.5. Eclipse with m2e doesn't nag either.

Ok to skip the check alltogether then?

agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
```diff
@@ -35,6 +40,27 @@
  */
 public final class RestrictorFactory {
 
+    private static final Pattern URL_PATTERN = Pattern.compile("(http[s]?://)((.+):(.+)@)?(.+)");
```
rhuss: I wonder whether, how to deal with 'user:' and ':pass'. I guess 'user:' is valid, when we use an empty password ? but ':pass' is probably really an error. 

So maybe we should allow `(?<protocol>http[s]?://)(?:(?<user>.+:)(?<password>.+)?@)?.*` (here with named capturing groups), so that empty passwords are also matched ?

BTW, what is with password containing `@` ? (don't know whether that is supported). however should work here since the regexp is greedy. Maybe you could add a test with an password containing a `@`, too ?

agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
```diff
@@ -115,9 +142,35 @@ public static PolicyRestrictor lookupPolicyRestrictor(String pLocation) throws I
...
+        Objects.requireNonNull(pLocation);
+
+        Matcher matcher = URL_PATTERN.matcher(pLocation);
+
+        String url = pLocation;
+        String basicAuthHeaderValue = null;
+
+        // We got a basic-auth user and password
+        if (matcher.matches() && matcher.group(3) != null && matcher.group(4) != null) {
+            url = matcher.group(1) + matcher.group(5);
```
rhuss: See above, you could also used named groups which often make things easier to understand

```java
url = matcher.group("protocol") + matcher.group("rest");
```

agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
mweirauch: Because it's printed into the logs and would contain - at least in the Basic-Auth case - sensitive information which you don't wan't spread out into your central logging system for public consumption.

But yap, I didn't think about the non-sensitive case. Namely status-quo before.  No quick solution in my head for the moment.

rhuss:
Looks good to me. We just need to decide what todo with an empty password, imo this is valid (but not 100% sure). And a test for a password containing a `@` would be good. In my understanding everything until the last `@` should be taken. However not really sure here. (because an url can contain @, too ?)

agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
```diff
@@ -115,9 +142,35 @@ public static PolicyRestrictor lookupPolicyRestrictor(String pLocation) throws I
                 is = RestrictorFactory.class.getResourceAsStream(path);
             }
         } else {
-            URL url = new URL(pLocation);
-            is = url.openStream();
+            URLConnectionResult urlConnectionResult = buildUrlConnection(pLocation);
```
rhuss: I wonder why you need this temporary result object anyway. Couldn't you simply return the UrlConnection itself ?

If it is only for the test, I would prefer another strategy. Put the parsing stuff in an utility class and test this since this is the only critical part here. Change the original code for better testing is a smell per se.

I mean you could have a static 'String[] extractUrlAndAuthentication(url)' method which returns both parts url, and encoded base64 string in a two element array. This could be then tested. Ok, not perfect, too, but you really dont need this result object imo.

mweirauch:
Thanks for the comments! Didn't even know named capturing groups exist ;)

An empty password seems - according to some discussion - valid. An empty user not.

Will try to address the issues.

agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
rhuss: Also this only works if you use properties to obfuscate the password. What about using a regexp as above and simply replace the password with '*' on the resolved property before logging it ?

agent/core/src/main/java/org/jolokia/restrictor/RestrictorFactory.java
mweirauch: Yeah, it resulted after I figured the Sun implementation for the URLConnection will not expose any "security headers" even if they are set. Extra method then as you describe.

rhuss:
@mweirauch Any news on this ? ;-)

mweirauch:
Eek. @rhuss unfortunately not. I got burried away with other stuff. And the worst of all I found a "solution" for providing the policyLocation basic-auth-less which is a little demotivator.

I can't yet promise to provide a cleaned up and less intrusive solution "in time".

rhuss:
np :) will probably have a look on this later on my own

thanks anyway !

grgrzybek:
I closed #276 with (I think) sufficient explanation. I'm closing this PR too.

mweirauch:
Thanks for taking care of this. All fine. I had long forgotten about this issue, which isn't one anymore.

PR #281, opened 2016-09-07
Better version detection of Payara Server

savage-engineer:
Currently Jolokia will detect Payara Server's version, but will still supply the product name as `glassfish` and the vendor name as `Oracle`. My proposed change separates the patterns out so that Payara Server is detected better.

The detector will check first for Glassfish and then for Payara Server.


davsclaus:
@rhuss looks good to me

rhuss:
Thanks !

Issue #291, opened 2016-11-01
Class cast exception when checking server mbean


robstryker:
!ENTRY org.eclipse.core.jobs 4 2 2016-11-01 16:02:45.695
!MESSAGE An internal error occurred during: "Connect Job".
!STACK 0
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Long
	at org.jolokia.client.request.ValidatingResponseExtractor.extract(ValidatingResponseExtractor.java:58)
	at org.jolokia.client.J4pClient.execute(J4pClient.java:195)
	at org.jolokia.client.J4pClient.execute(J4pClient.java:168)
	at org.jolokia.client.J4pClient.execute(J4pClient.java:117)
	at org.jboss.tools.jmx.jolokia.JolokiaConnectionWrapper.verifyServerReachable(JolokiaConnectionWrapper.java:318)

This occurs when pJsonResponse is:
{"reason":"Forbidden","metadata":{},"apiVersion":"v1","code":403,"kind":"Status","details":{"kind":"pods\/proxy","name":"https:eap-app-6-9t61s:8778"},"message":"User \"openshift-dev\" cannot get pods\/proxy in project \"projName\"","status":"Failure"}

grgrzybek:
Fixed in 5671228e

PR #292, opened 2016-11-01
#291 - avoid class cast exception when server is forbidden

robstryker:
I'm not able to build this locally, but the fix seems simple enough. :|  Please feel free to verify before accepting the patch.

### DIFF
--- a/client/java/src/main/java/org/jolokia/client/request/ValidatingResponseExtractor.java
+++ b/client/java/src/main/java/org/jolokia/client/request/ValidatingResponseExtractor.java
@@ -54,10 +54,14 @@ public class ValidatingResponseExtractor implements J4pResponseExtractor {
     public <RESP extends J4pResponse<REQ>, REQ extends J4pRequest> RESP extract(REQ pRequest,
                                                                       JSONObject pJsonResp)
             throws J4pRemoteException {
-
-        int status = pJsonResp.containsKey("status") ?
-                ((Long) pJsonResp.get("status")).intValue() :
-                0;
+	
+	int status = 0;
+	if( pJsonResp.containsKey("status")) {
+		Object o = pJsonResp.get("status");
+		if( o instanceof Long ) {
+			status = ((Long)o).intValue();
+		} 
+	}
 
         if (!allowedCodes.contains(status)) {
             throw new J4pRemoteException(pRequest, pJsonResp);
--- a/client/java/src/main/java/org/jolokia/client/exception/J4pRemoteException.java
+++ b/client/java/src/main/java/org/jolokia/client/exception/J4pRemoteException.java
@@ -60,16 +60,25 @@ public class J4pRemoteException extends J4pException {
     }
 
     public J4pRemoteException(J4pRequest pJ4pRequest, JSONObject pJsonRespObject) {
-        super(pJsonRespObject.get("error") != null ?
-                      (String) pJsonRespObject.get("error") :
-                      "Invalid response received: " + pJsonRespObject.toJSONString()
-             );
-        Long statusL = (Long) pJsonRespObject.get("status");
+        super(generateErrorMessage(pJ4pRequest, pJsonRespObject));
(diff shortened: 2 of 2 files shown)


apupier:
the error of the job is
{noformat}
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.2:compile (default-compile) on project jolokia-client-java: Compilation failure
[ERROR] /home/travis/build/rhuss/jolokia/client/java/src/main/java/org/jolokia/client/request/ValidatingResponseExtractor.java:[59,44] ')' expected
{noformat}

client/java/src/main/java/org/jolokia/client/request/ValidatingResponseExtractor.java
```diff
@@ -54,10 +54,12 @@ public ValidatingResponseExtractor(int... pCodesAllowed) {
     public <RESP extends J4pResponse<REQ>, REQ extends J4pRequest> RESP extract(REQ pRequest,
                                                                       JSONObject pJsonResp)
             throws J4pRemoteException {
-
-        int status = pJsonResp.containsKey("status") ?
-                ((Long) pJsonResp.get("status")).intValue() :
-                0;
+	
+	int status = 0;
+	if( pJsonResp.containsKey("status") {
```
rhuss: You miss a closing ')' at the end of the "if" clase)

rhuss:
Sorry @robstryker but we can't include the pull request like it is in upstream Jolokia. The reason is that this is very specific to the OpenShift API server and although the API server returns a JSON object, too, Jolokia has no a priori knowledge about this format. It's really very specific to the OpenShift setup and we must not couple the (generic) Jolokia Java client to this. The 'code' which you are using as a fallback is not part of the Jolokia protocol, it's just from a completely different protocol. 

The only think I'm afraid what we can do here is to be more passive and check for the status type. It's just a coincidence that OpenShift in its API and Jolokia in its API uses the same field. Actually, I would even return an error if there is a type mismatch as its clearly not a proper Jolokia response.

robstryker:
Thats fine... at least we can avoid a RuntimeException via the class cast error, even if we don't use the 'code' fallback?    A J4pRemoteException is much better than a runtime exception, no?

robstryker:
I updated the patch to get rid of the 'else'.  At least then we avoid the class cast error.

client/java/src/main/java/org/jolokia/client/request/ValidatingResponseExtractor.java
```diff
@@ -54,10 +54,14 @@ public ValidatingResponseExtractor(int... pCodesAllowed) {
...
-        int status = pJsonResp.containsKey("status") ?
-                ((Long) pJsonResp.get("status")).intValue() :
-                0;
+	
+	int status = 0;
+	if( pJsonResp.containsKey("status")) {
+		Object o = pJsonResp.get("status");
+		if( o instanceof Long ) {
+			status = ((Long)o).intValue();
+		}
```
rhuss: The question is still what we should when status in **not** a long. I think its better to throw a `J4pRemoteException` at this case (including the value and type of the status) instead of silently ignoring it.

@robstryker wdyt ?

robstryker: As currently coded, it isn't going to be silently ignored. The int status will maintain it's 0 value, which will cause the J4pRemoteException below to be thrown, since 0 is not a valid response code in allowedCodes. 

So I think the current patch is correct, but if you wanted to be explicit and duplicate the throw, then I'm sure that's fine by me too.

rhuss: Yes, you are right, but I think its better to include the extra context information which type and (string) value status currently have, like ("Invalid status of type java.lang.String ('error') received"). The more context info in an error, the better.

robstryker:
@rhuss  Any chance we can get this in a build?

PR #307, opened 2017-02-23
Adds debian package generation on the jvm agent

xiu:
Debian package will be generated on "mvn package"


rhuss:
Thanks ! 

@xiu So now that we have that in place the next step is how to add it to the release process. I.e. where to upload the .deb. Should we upload it as an artefact with a classifer to Maven central, too ? Or is there a public Debian repository which could be used ? 

I then would move this into a profile which is activated only during a release as it is only necessary for a release (and slows down the build otherwiser)

rhuss:
Thanks ! Sorry for the delay.

I decided to upload the deb packages directly to github starting with the next release.

PR #316, opened 2017-03-30
Fix various typo in documentation

apupier:


rhuss:
Awesome, thanks a lot ! Finally someone is reading the docs ;-) No, just kidding, its really awesome to get that kind of feedback which helps improving documentation (which is always a very important aspect for every piece of software)

merci bien :)

Issue #317, opened 2017-03-30
Provide classname from MBeanInfo


apupier:

apupier:
fixed in 1.3.6

PR #318, opened 2017-03-30
#317 Provide classname MBeanInfo on List request

apupier:
Signed-off-by: Aurelien Pupier <apupier@redhat.com>

### DIFF
--- a/agent/core/src/test/java/org/jolokia/handler/ListHandlerTest.java
+++ b/agent/core/src/test/java/org/jolokia/handler/ListHandlerTest.java
@@ -30,6 +30,7 @@ import org.jolokia.request.JmxListRequest;
 import org.jolokia.request.JmxRequestBuilder;
 import org.jolokia.restrictor.AllowAllRestrictor;
 import org.jolokia.config.ConfigKey;
+import org.jolokia.handler.list.DataKeys;
 import org.jolokia.util.RequestType;
 import org.testng.annotations.*;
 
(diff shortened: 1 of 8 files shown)


agent/core/src/test/java/org/jolokia/handler/ListHandlerTest.java
```diff
@@ -168,7 +168,7 @@ public void truncatedList() throws Exception {
         JmxListRequest request = new JmxRequestBuilder(RequestType.LIST).pathParts("java.lang", "type=Runtime").build();
         Map res = execute(request);
         assertFalse(res.containsKey("op"));
-        assertEquals(res.size(),2);
+        assertEquals(res.size(),3);
```
apupier: @rhuss for this one I'm not completely sure if it is coherent

apupier: and might need to update https://github.com/rhuss/jolokia/blob/b6010e3c053648eb280e57c1255f741d579b0688/src/docbkx/protocol/list.xml#L158 then, will do it tomorrow if your answer is positive

rhuss: Should be fine. The only key missing here is "op" as the RuntimeMBean does not have an operations.

Would be cool if you could update the docs, too (and maybe replace "Reduce the depth of depth" by "Reduce the depth of the tree" while you are at it ;-)

apupier: done

rhuss:
Thanks, looks good !

PR #323, opened 2017-04-25
Upgrade httpclient-osgi 4.3.3 -> 4.3.6

pgier:
Note, starting in 4.3.6, the httpcore classes are no longer bundled
with httpclient-osgi, so a separate dependency declaration is required.
See https://issues.apache.org/jira/browse/HTTPCLIENT-1558

### DIFF
--- a/client/java/pom.xml
+++ b/client/java/pom.xml
@@ -41,7 +41,7 @@
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpcore</artifactId>
-      <version>4.4.6</version>
+      <version>4.4.7</version>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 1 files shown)


rhuss:
Thanks for the heads up. Strange to see such a breaking change in a patch level version of httpclient, though.

will hopefully have time to review and intergrate this PR soon.

Thanks !

pgier:
Yeah, I was surprised to see that change in 4.3.6 as well.  Another way this upgrade could be done is to remove the transitive dep exclusions from httpclient-osgi, and then httpcore should be picked up transitively instead of requiring the direct dep on httpcore.

rhuss:
thanks again, and sorry for the long delay !

PR #326, opened 2017-05-31
Issue 325 - unify error messaging in J4pRemoteException

robstryker:


robstryker:
I've updated this PR.

PR #331, opened 2017-07-23
Spring backend

skarsaune:
I have been testing the spring agent in some applications. Here are a couple of showstoppers that I have encountered:
* Allow jolokia to fall back to other handlers for requests not handled by the spring handler without throwing a runtime exception.
* Tolerate Spring beans that may not have a class name, such as the context


rhuss:
Thanks a lot ! 

And especially for testing out the spring handler, which has not been used much in the wild yet. Feel free to give feedback as you go.

Unfortunately, Jolokia 2.0 has not gained much traction in the last year, and I have the impression that there is not really much demand over 1.0 yet, so other tasks bubbled up in my (never ending) backlog. I stop now with 'promises' about when 2.0 will be GA, but I will continue to make further milestones.

PR #332, opened 2017-07-30
Spring list meta

skarsaune:
…method and kind of data

Signed-off-by: marska <martin.skarsaune@kantega.no>

Rationale: No longer test for read method, as it seems that it is taken for granted and dealt with via field reflection in the read handler. Initially make writeability limited by datatype. 
 Avoid abstract beans as they cannot be retrieved from the context and interacted with. 

Background: Not really thorough testing in the wild, but playing around with in on a few client projects. From the looks of things there are a lot of variations possible and we have probably just scratched the surface......


rhuss:
Looks good ! For the write meta data I wonder, whether we should reflect whether a property is writable per-se or whether writable via Jolokia (which is more limited because of restricted serialization caps). I think its already clear from the protocol description that a writable property cannot be updated via the Jolokia protocol if it's a complex data-type, so I would like to keep the property indepedendent of the Jolokia proto caps (or in other words I think the check on a literal value should not happen on that level, but on the client libs).

skarsaune:
Good point. Have removed the requirement that it must be a literal type.

rhuss:
Just rebased and merged. Thanks ! (sorry for the delay)

PR #342, opened 2017-09-05
Several modifications to make the Jolokia JVM Agent start and work on…

danpfe:
Hi Roland,
currently at a customer that really wants to use the IBM JVM for some of their Docker-images and thus I had to modify the Jolokia JVM Agent that I want to use for my projects to work with that JVM as well. Hope it's useful. In short, Jolokia seems to work for me now and the tests go through on both OpenJDK 8 and IBM J9 v8.
/D


danpfe:
Oops, just saw that "All checks have failed" ( what an exaggeration ;-) ), it seems the Oracle JDK7 checks fail due to:

```
Error: JAVA_HOME is not defined correctly.
  We cannot execute /usr/lib/jvm/java-7-oracle/bin/java
```

danpfe:
Forgot to add the contribution sign-off. Sending a new pull request.

PR #343, opened 2017-09-06
Modifications to make JVM Agent work on IBM JVM 8.

danpfe:
Mods to make JVM Agent work on the IBM JVM 8. Hopefully this is useful for those that happen to run IBM's JDK and want to use the Jolokia JVM Agent.


rhuss:
Thanks a lot ! looks good on a first sight. Unfortunately I'm a bit busy these days, but I'm going to review and integrate this PR asap. Please ping me, if I don't come back to this within the next two weeks.

thanks again ...

agent/jvm/src/main/java/org/jolokia/jvmagent/security/KeyStoreUtil.java
```diff
@@ -137,24 +147,48 @@ private static Object createKeyPair() throws NoSuchAlgorithmException, NoSuchPro
...
+    private static X509Certificate getSelfCertificate(Object keypair, Object x500Name, Date date, long l) {
+        final Class<?> clazz;
+        if (ClassUtil.checkForClass(X500_NAME_SUN)) {
+            clazz = ClassUtil.classForName(X500_NAME_SUN, false);
+        } else {
+            clazz = ClassUtil.classForName(X500_NAME_IBM, false);
+        }
+
+        try {
+            final Method selfCertMethod = keypair.getClass().getDeclaredMethod("getSelfCertificate", clazz, Date.class, long.class);
```
rhuss: I wonder, whether the original `ClassUtil.applyMethod()` wouldn't work here as well ? The incoming `x500Name` object already would be already of the proper IBM / Oracle type and `applyMethod()` should work in both cases. Have you tried it and got some error ?

agent/jvm/src/main/java/org/jolokia/jvmagent/security/KeyStoreUtil.java
```diff
@@ -137,24 +147,48 @@ private static Object createKeyPair() throws NoSuchAlgorithmException, NoSuchPro
...
     private static PrivateKey getPrivateKey(Object keypair) {
         return (PrivateKey) ClassUtil.applyMethod(keypair, "getPrivateKey");
     }
 
     private static Class lookupKeyGenClass() {
-        Class keyGenClass = ClassUtil.classForName(KEYGEN_CLASS_JDK8);
-        if (keyGenClass == null) {
-            keyGenClass = ClassUtil.classForName(KEYGEN_CLASS_JDK7);
+        final Class keyGenClass;
+        if (ClassUtil.checkForClass(KEYGEN_CLASS_JDK8_SUN)) {
```
rhuss: Since `classForName` already return `null` when the class could not be found we could save all the calls to `checkForClass()`: Just fetch the class and if null, try next. This could be even moved out into an extra method which makes this in a loop and returns as soon as the first has been found.

agent/jvm/src/test/java/org/jolokia/jvmagent/client/util/VirtualMachineHandlerTest.java
```diff
@@ -56,6 +56,11 @@ public void emptyPid() {
 
     @Test
     public void listAndAttach() throws Exception, NoSuchMethodException, IllegalAccessException {
+        // FIXME Fails on the IBM JVM; however, the agent seems to work. Skip for now just to get the tests to run on the IBM JVM. - Anyone?
```
rhuss: Any idea why and where this fails ?

agent/jvm/src/test/java/org/jolokia/jvmagent/JolokiaServerTest.java
```diff
@@ -294,7 +293,20 @@ public void sslWithSpecialHttpsSettings() throws Exception {
 
                 try {
                     TrustManager tms[] = getTrustManagers(true);
-                    SSLContext sc = SSLContext.getInstance(protocol);
+                    SSLContext sc;
+                    try {
+                        sc = SSLContext.getInstance(protocol);
+                    } catch (NoSuchAlgorithmException e) {
+                        // on the IBM JVM v8 we get an exception on SSLv3 which fails the entire test, safe-guarding for it.
```
rhuss: Any idea why this fails for IBM ? If this is inevitable I would check also for the IBM VM and only continue if its on IBM (and fail like before for Oracle).

agent/jvm/pom.xml
```diff
@@ -184,6 +208,30 @@
...
+        <jvm.algorithm>IBMX509</jvm.algorithm>
+      </properties>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <forkMode>always</forkMode>
+              <argLine>-javaagent:${settings.localRepository}/org/jmockit/jmockit/1.8/jmockit-1.8.jar</argLine>
```
rhuss: would be cool when if we could reuse the same jmockit version as declared in the dependency. Maybe extract it into a property which then is referenced here and within the dependency ?

agent/jvm/src/main/java/org/jolokia/jvmagent/security/KeyStoreUtil.java
danpfe: Yes, it fails because the applyMethod would search for a method-signature that does not exist (if I remember correctly right now it'd say that there is no method that takes Object, Date, long). Of course it would be possible to modify applyMethod but at that moment I thought it's less invasive in your code to simply do it this way.

agent/jvm/src/main/java/org/jolokia/jvmagent/security/KeyStoreUtil.java
danpfe: Yes, that wouldn't be a problem, at the moment I wrote the code it just didn't look very pleasing to my eye to see if this null, if that null, etc :) I can change it if you like.

agent/jvm/src/test/java/org/jolokia/jvmagent/client/util/VirtualMachineHandlerTest.java
danpfe: Where it fails? Directly at the first assert (when it checks if the process-collection is not empty). Why it fails I am honestly not sure about (as I myself very very rarely use J9). However, my theory here is that we are running into the same "lack of attach-mode"-problem as with JMockit. I know the class exists otherwise it would fail earlier and when I tried to find a reason for this I did a fair bit of googling and it seems that the general consensus in that respect is that you simply can't attach to J9 the same way you can with OpenJDK/Oracle JDK. As I wrote in the comment, Jolokia itself seems to work just fine after a fair amount of manual testing but it's naturally a bit disturbing to have to disable a test. I could look into it once more during the weekend, perhaps I'll find the root-cause.

agent/jvm/src/test/java/org/jolokia/jvmagent/JolokiaServerTest.java
danpfe: Yeah, now that I look at it again I believe this should be changed (partially because the reason given was what I thought in the beginning to be the reason but later forgot to change it before sending the pull request). From what I found out SSLv3 is simply disabled, which is a bit confusing because the list of cipher suites all start with SSL_, but what's TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA on OpenJDK is SSL_ECDHE_RSA_WITH_AES_128_CBC_SHA on IBM.

agent/jvm/pom.xml
danpfe: Correct, my bad. I intended to change it before committing but this bad boy slipped through :)

agent/jvm/src/test/java/org/jolokia/jvmagent/client/util/VirtualMachineHandlerTest.java
danpfe: Not much for my theory it seemed. "attach" worked fine, "detach" failed with IllegalAccessException, sent another commit your way. Think all your requested changes are resolved now.

rhuss:
seems to be a minor merge conflict. would be also awesome if you could resolve this by rebasing to the current master.

Thanks!

danpfe:
Sure, will fix tonight.

danpfe:
Rebased, fixed conflict, squashed commits, pushed. Hope this is useful for those few of us that from time to time are on J9.

rhuss:
Thanks a lot ! 'hope I can get a version out soon, but very likely not this week anymore.

PR #347, opened 2017-09-11
Adding LightStreamer Detector

sandy-ig:
- LightStreamer starts its own MBean server for internal metrics
- The Jolokia plugin is intermittent in picking the MBeans up due to startup timing issues between the JavaAgent and the MBean Server
- The detector will look for System Properties or Classes which belong to LightStreamer and if present delay the MBean scanning until the LightStreamer MBean classes have been loaded

Signed-off-by: Alexander Gordon <alexander.gordon@ig.com>

### DIFF
--- a/agent/core/src/main/java/org/jolokia/detector/LightstreamerDetector.java
+++ b/agent/core/src/main/java/org/jolokia/detector/LightstreamerDetector.java
@@ -1,13 +1,22 @@
 package org.jolokia.detector;
 
 import java.lang.instrument.Instrumentation;
+import java.util.Properties;
 import org.jolokia.backend.executor.MBeanServerExecutor;
 
 public class LightstreamerDetector extends AbstractServerDetector {
 
-
-    /** {@inheritDoc}
-     * @param pMBeanServerExecutor*/
+    private static final String[] SYSTEM_PROERTY_NAMES = new String[]{
+        "com.lightstreamer.internal_lib_path",
+        "com.lightstreamer.kernel_lib_path",
+        "com.lightstreamer.logging_lib_path",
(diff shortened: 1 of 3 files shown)


agent/core/src/main/java/org/jolokia/detector/LightstreamerDetector.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+    
+    protected boolean isLightStreamer(Instrumentation instrumentation) {
+        for (String name : System.getProperties().stringPropertyNames()) {
+            if (name.contains("com.lightstreamer")) {
+                return true;
+            }
+        }
+        for (Class loadedClass : instrumentation.getAllLoadedClasses()) {
```
rhuss: Isn't this a bit heavy weight to iterate over all loaded classes ? This will slow down every startup even when not running on lightstream.

agent/core/src/main/java/org/jolokia/detector/LightstreamerDetector.java
```diff
@@ -0,0 +1,68 @@
...
+            if (loadedClass.getCanonicalName().contains("com.lightstreamer")) {
+                return true;
+            }
+        }
+        return false;
+    }
+    
+    public static final int LIGHTSTREAMER_DETECT_TIMEOUT = 5 * 60 * 1000;
```
rhuss: guess we can keep the variables private, they are only use within this class.

sandy-ig:
Have updated the PR with regards to your comments.
It now targets 1 of 4 possible System Properties which should always be present to identify a LightStreamer system and removed the Class searching as this was just an additional backup search.
Variables have been made private as well

PR #353, opened 2017-10-03
Give discovery threads names to indicate ownership

skarsaune:
Signed-off-by: marska <martin.skarsaune@kantega.no>

The muliticast socket listener (discovery) threads have generic names Thread-n . As jolokia runs as an agent it should mark ownership of these threads to avoid unneccesary confusion:
image

Suggestion: prefix threads with JolokiaDiscoveryListenerThread to state what they are:
image

There are still 2 generically named threads from com.sun.net.httpserver, but there does not seem to be any good way to affect this as the thread is instantiated inside the server source.


rhuss:
Thanks !

PR #371, opened 2018-01-23
fix(json streaming): Removed ChunkedWriter

rhuss:
and add IoUtil.streamJsonAndClose() as an alternative (+ test). 

@gnufied @mtaylor Could you please have a brief look, if this PR makes sense ? I'm just running the integration test suite, but it looks good so far. From my understanding this code only differs in the buffer size of the original ChunkedWriter (8192 vs. 4096)


rhuss:
I did integration tests for all of the supported agents (on tomcat, wildfly and felix), and it looks good.

@gnufied @mtaylor So if you don't mind I'm going with this solution and prepare a Jolokia release 1.4.0 out of the box.

mtaylor:
@rhuss +1. Looks good to me.  Cheers

gnufied:
As I discussed this in IRC with @mtaylor yesterday, we should also test this externally with a running application using an external client like `urllib2` of Python or something similar. We can run the test in a close loop to make sure, we are not adding any regressions.

rhuss:
@mtaylor cool thanks ;-)

rhuss:
@gnufied I use this test suite --> https://github.com/rhuss/jmx4perl/tree/master/it which has around 300 integration tests around various aspects of Jolokia. It uses the Perl HTTP module to connect. 

I ususally run this against 25 different servers, for now I did it for 

```
Jolokia 1.4.0
   Java EE Server
      Tomcat 7.0.52
   JVM
      Wildfly 10.1.0.Final
   OSGi
      Felix 4.2.1 (pure & bundle)
```

which is a representative selection for now (tbh, the original streaming implementation has also not been tested on much more platforms)

rhuss:
So I'm think we are good to merge. Streaming can be always switched off if there should be issues (although it is switched on by default now).

@mtaylor @gnufied Thanks a lot for your support and help !

I'm going to release Jolokia 1.4.0 right now ...

PR #372, opened 2018-03-07
Slf4j LogHandler implementation

ags313:
Slf4j Log Handler

### DIFF
--- a/agent/core/pom.xml
+++ b/agent/core/pom.xml
@@ -49,6 +49,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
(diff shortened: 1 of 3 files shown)


rhuss:
Thanks for the PR !

However, there is a reason why Jolokia not yet switched to any given, external logging framework: The number of dependencies should be kept minimal and in fact Jolokia has only a single dependency, json-simple. That helped in keeping the agent focus and small (~450k). Addint sl4j API would increase the overall size already by ~ 10%.

That said, you can of course create a customized agent using an LogHandler implementation based on another logging API, its just that I'm a bit reluctant to include this in the core distribution.

However, what we could consider to introduce java.util.Logging, as we are giving up on support for Java 1.6 recently. Would this help in your use case ?


And for Jolokia 2 we could consider a different approach alltogether.

ags313:
I have my own LogHandler already. 
"provided" scope does not increase the weight, and having an sl4fj entry point reduces entry cost.

Issue #334, opened 2017-07-31
Support BigInteger as return type for method Invocation


apupier:
it is currently returning a JSonOBject with value {"lowestSetBit"=1L}

chemeugene:
+1

rhuss:
Agreed. A bit busy these, happy to take pull requests.

rhuss:
Its about writing a simplifier like the UrlSimplifier and register it in "META-INF/simplifiers-default"

apupier:
sounds like it was implemented a while ago.

closing

PR #383, opened 2018-06-28
Support BigInteger as return type for method Invocation #334

nevenr:


agent/core/src/test/java/org/jolokia/converter/json/SimplifiersTest.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Stack;
+
+import static org.testng.Assert.assertEquals;
+
+/**
+ * @author Neven Radovanović
+ * @since June 27, 2018
+ */
+@Test()
+public class SimplifiersTest {
```
rhuss: Thanks for the tests !

rhuss:
@nevenr Could you please resolve this (simple) conflict and maybe be squash your commits to a single one ? 

thanks !

rhuss:
merci !

Issue #367, opened 2018-01-06
Customizable thread names for JolokiaServer/JolokiaServerConfig


mjustin:
It would be useful to be able to configure the thread names of threads created by the Jolokia server.  That way if the application is logging thread names, it'll be obvious which threads were from Jolokia server requests.  For instance, in my app I might want to have the threads named "jolokia-1", "jolokia-2", ... instead of the Java default "Thread-1", "Thread-2", ...

This could be solved by adding a "threadNamePrefix" or "threadNameFormat" option to the JolokiaServerConfig.

Alternatively, if there was a way to specify the Executor or ThreadFactory used by the the HttpServer, the user could customize it to set whatever name they wanted.

The current workaround to achieve this is to create a JolokiaServer with a manually created HttpServer, rather than letting Jolokia create the server.  In addition to the additional code and potential inconsistency of manually configuring an HttpServer, it also means that it cannot be used with jolokia-spring's [SpringJolokiaAgent](https://github.com/jolokia-org/jolokia-spring/blob/master/src/main/java/org/jolokia/support/spring/SpringJolokiaAgent.java), as that uses the default server created by JolokiaServer.

Example of APIs in other frameworks that have this sort of behavior:
* Spring's [SimpleAsyncTaskExecutor](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/core/task/SimpleAsyncTaskExecutor.html#SimpleAsyncTaskExecutor-java.lang.String-).
* Guava's [ThreadFactoryBuilder](https://google.github.io/guava/releases/23.6-jre/api/docs/com/google/common/util/concurrent/ThreadFactoryBuilder.html#setNameFormat-java.lang.String-)

rhuss:
Sound like a very good idea. Fancy for doing a PR ? (i'm a bit loaded these days)

mjustin:
If I do just one of those options, I'm thinking "threadNamePrefix" as it's the most straightforward to implement. Sound reasonable?

rhuss:
"threadNamePrefix" sounds fine to me.

mjustin:
I have since switched to a different project that does not use Jolokia.  It will therefore be unlikely that I will find the time to make this improvement in the near future.

mjustin:
@rhuss — Looks like this was implemented back in 2018 and thus can be closed, correct?

rhuss:
yep. thanks for the heads up, stay safe ...

PR #384, opened 2018-06-30
Customizable thread names for JolokiaServer/JolokiaServerConfig #367

nevenr:
Add configuration option threadNamePrefix which will be used by jolokia http server executor 
Default: jolokia-


rhuss:
thanks, its fine for me. If the increment is synchronized, that this should be no problem.

Issue #344, opened 2017-09-07
JVM agent authentication is hardcoded to Mode.ANY


wsargent:
In JolokiaServerConfig, if HTTPS is on and client certificates are in use, then Jolokia treats this as an authentication of the JAAS subject:

```java
        if (useHttps() && useSslClientAuthentication()) {
            authenticators.add(new ClientCertAuthenticator(this));
        }
```

https://github.com/rhuss/jolokia/blob/master/agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServerConfig.java#L424

and likewise for JAAS (username and password):

```java
else if ("jaas".equalsIgnoreCase(authMode)) {
            authenticators.add(new JaasAuthenticator(realm));
        }
```

https://github.com/rhuss/jolokia/blob/master/agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServerConfig.java#L436

and the MultiAuthenticator is enabled for ANY inside a private method:

```java
            authenticator = new MultiAuthenticator(MultiAuthenticator.Mode.ANY, authenticators);
```

https://github.com/rhuss/jolokia/blob/master/agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServerConfig.java#L454

This is a problem, because I'm trying to get multi-factor authentication working with Jolokia, so I want a JAAS module AND a TLS client certificate, which I can do with JConsole / JMX / RMI with [Yubikey JAAS module](https://github.com/Yubico/yubico-java-client/tree/master/jaas):

```
-Dcom.sun.management.jmxremote.registry.ssl=true \
-Dcom.sun.management.jmxremote.ssl=true \
-Dcom.sun.management.jmxremote.authenticate=true \
-Djava.security.auth.login.config=$CONFDIR/yubikey.config \
-Dcom.sun.management.jmxremote.login.config=YubikeyAuth \
-Dcom.sun.management.jmxremote.ssl.need.client.auth=true 
```

So it's a bit odd not to be able to do it in Jolokia.

rhuss:
You are right, this sounds like a valid use case. 

I'm currently quite busy these days, but the implementation should be straightforward by introducing a configuration key `authPass`, with default `any` (for backwards compatibility) and alternate value `every`.

I would happily review and integrate any pull request which adds this feature (including doc & changelog update, and unit test if not too involved).

grgrzybek:
[...]

PR #385, opened 2018-07-01
 JVM agent authentication is hardcoded to Mode.ANY #344 

nevenr:


agent/core/src/main/java/org/jolokia/config/ConfigKey.java
```diff
@@ -222,6 +223,12 @@
      */
     AUTH_MODE("authMode", true, false, "basic"),
 
+    /**
+     * If MultiAuthenticator is used, this config item explains how to combine multiple authenticators
+     * Supported values: "any" at least one authenticator must match, "every" or "all" all authenticators must match
+     */
+    AUTH_PASS("authPass",true, false, "any"),
```
rhuss: can we ommit 'every' and use only 'all' ? Doesn't make much sense to keep both and only increases the maintenance burden.

agent/jvm/src/main/java/org/jolokia/jvmagent/security/MultiAuthenticator.java
```diff
@@ -48,8 +49,22 @@
...
-        ANY
-    };
+        ANY;
+
+        public static Mode fromString(String inStr) {
+            if (inStr == null || inStr.isEmpty()) {
+                throw new IllegalArgumentException("String representation of MultiAuthenticator.Mode can not be null nor empty.");
+            }
+            inStr = inStr.toLowerCase();
+            if (inStr.equals("any")){
```
rhuss: could be also just `.equalsIgnoreCase()`

src/changes/changes.xml
```diff
@@ -23,6 +23,12 @@
     <author email="roland@jolokia.org">Roland Huß</author>
   </properties>
   <body>
+    <release version="1.6.1-SNAPSHOT" description="Release 1.6.1-SNAPSHOT" date="yyyy-mm-dd">
+      <action dev="nevenr" type="add">
```
rhuss: just add a `issue=""` attribute here, so you don't have to reference it in the comment (see other entries for examples)

agent/core/src/main/java/org/jolokia/config/ConfigKey.java
rhuss: Maybe we rename it to `authMatch` instead of `authPass` ? Not sure, sounds a better fit to me ...

rhuss:
Thanks a lot !

agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServerConfig.java
```diff
@@ -451,7 +451,7 @@ private void initAuthenticatorFromAuthMode() {
         } else {
             // Multiple auth strategies were configured, pass auth if any of them
             // succeed.
-            authenticator = new MultiAuthenticator(MultiAuthenticator.Mode.ANY, authenticators);
+            authenticator = new MultiAuthenticator(MultiAuthenticator.Mode.fromString(jolokiaConfig.get(ConfigKey.AUTH_MATCH)), authenticators);
```
rhuss: btw, if the config is null an error is thrown by fromString. I changed this to return ANY for a none-config (as this was the default before).

PR #390, opened 2018-09-05
chore: Remove accidental references to non Java 1.6 classes

### LINKED ISSUE #388: Jolokia uses JRE 1.8 classes while supporting 1.6
Importing sources into Eclipse, it starts complaining about missing `java.beans.Transient` classes.
This is correct as that is not part of JRE 1.6, only of 1.8.

Jolokia 1.6.0. therefore does not support JRE 1.6.

rhuss:
Fixes #388.


G-Ork:
With maven toolchains this would fail at compile time.

https://maven.apache.org/guides/mini/guide-using-toolchains.html

PR #404, opened 2019-05-23
fix: Add support for multiple CA entries

### LINKED ISSUE #403: Add support to load multiple CA certs from store
It'd be great to add a support for multiple CA certs. 

https://github.com/rhuss/jolokia/blob/33b23e578e6e9dbb0604e3168732ceb332cbba9a/agent/jvm/src/main/java/org/jolokia/jvmagent/security/KeyStoreUtil.java#L55-L67

In fact it's a bug affecting latest versions of OpenShift. The same issue was fixed in openshift-ping. 
https://github.com/jboss-openshift/openshift-ping/pull/43

Cc @rhuss

dsimansk:
Fix #403


agent/jvm/src/test/java/org/jolokia/jvmagent/security/KeyStoreUtilTest.java
```diff
@@ -59,6 +59,33 @@ public void testTrustStore() throws KeyStoreException, CertificateException, NoS
...
+        KeyStore keystore = createKeyStore();
+
+        KeyStoreUtil.updateWithCaPem(keystore, caPem);
+
+        Enumeration<String> aliases = keystore.aliases();
+        String alias = aliases.nextElement();
+        assertTrue(aliases.hasMoreElements());
+        assertTrue(alias.contains("another.test.jolokia.org"));
+        X509Certificate cert = (X509Certificate) keystore.getCertificate(alias);
+        cert.checkValidity();
```
rhusar: This seems unnecessary and rather incorrect. The validity check `java.security.cert.X509Certificate#checkValidity()` throws `java.security.cert.CertificateExpiredException` so this test is bound to start failing in year 2025.

agent/jvm/src/test/java/org/jolokia/jvmagent/security/KeyStoreUtilTest.java
```diff
@@ -59,6 +59,33 @@ public void testTrustStore() throws KeyStoreException, CertificateException, NoS
...
+    public void testTrustStoreWithMultipeEntries() throws KeyStoreException, CertificateException, NoSuchAlgorithmException, IOException {
+        File caPem = getTempFile("ca/cert-multi.pem");
+        KeyStore keystore = createKeyStore();
+
+        KeyStoreUtil.updateWithCaPem(keystore, caPem);
+
+        Enumeration<String> aliases = keystore.aliases();
+        String alias = aliases.nextElement();
+        assertTrue(aliases.hasMoreElements());
+        assertTrue(alias.contains("another.test.jolokia.org"));
```
rhusar: It's an incorrect assumption of the test to require an `Enumeration` to be deterministically ordered. The contract of Enumeration does not mandate that (https://docs.oracle.com/javase/7/docs/api/java/util/Enumeration.html).

agent/jvm/src/test/java/org/jolokia/jvmagent/security/KeyStoreUtilTest.java
rhuss: Agreed, we should probably only check whether the keystore can be parsed and has multiple elements.

agent/jvm/src/test/java/org/jolokia/jvmagent/security/KeyStoreUtilTest.java
rhuss: Agreed. Instead it should be iterated over all elements and check whether  all elements are included (in any order).

dsimansk: Sorry, I'll refactor the test. I was a little hash doing the PR.

agent/jvm/src/test/java/org/jolokia/jvmagent/security/KeyStoreUtilTest.java
dsimansk: This's also true for former tests. Let me know if I should address that as well or a separate issue/PR is required.

agent/jvm/src/test/java/org/jolokia/jvmagent/security/KeyStoreUtilTest.java
rhuss: @rhusar thanks for the heads up ! I fixed the tests on master, should be fine now (no checkValidity(), don't imply order in enumerations)

rhuss: @dsimansk I just pushed an update. That might already addresses that issue. Could you please have a brief look ?

rhusar: @rhuss  @dsimansk  Looks good now!

PR #420, opened 2019-10-15
Also set secure processing feature for the PolicyRestrictor

coheigea:
We should also set the secure processing feature for the PolicyRestrictor, to prevent various DoS style attacks.


rhuss:
Thanks !

Issue #421, opened 2019-10-22
Make multicast discovery address configurable


graben:
Multicast address is actually hard coded to:
```
// IPv4 Address for Jolokia's Multicast group
public static final String JOLOKIA_MULTICAST_GROUP = "239.192.48.84";

// Multicast port where to listen for queries
public static final int JOLOKIA_MULTICAST_PORT = 24884;
```

This is problematic in large networks where discovery should be separated into clusters. Therefore, it would be helpful to explicitly config multicast address to agent.

rhuss:
Fair point. Fancy to work on a PR ? 'would help to integrate it (but otherwise totally busy with day-to-day work)

graben:
I will look to find some time for a PR.

PR #422, opened 2019-10-24
Add support for individual multicast address

graben:
@rhuss : PR for #421

### DIFF
--- a/agent/core/src/main/java/org/jolokia/config/ConfigKey.java
+++ b/agent/core/src/main/java/org/jolokia/config/ConfigKey.java
@@ -197,6 +197,16 @@ public enum ConfigKey {
      */
     DISCOVERY_AGENT_URL("discoveryAgentUrl",true,false),
 
+    /**
+     * IPv4 Address for Jolokia's Multicast group.
+     */
+    MULTICAST_GROUP("multicastGroup",true,false,"239.192.48.84"),
+
+    /**
+     * Multicast port where to listen for queries.
+     */
+    MULTICAST_PORT("multicastPort",true,false,"24884"),
+
     // ================================================================================
     // Configuration relevant for OSGI container
 
--- a/agent/core/src/main/java/org/jolokia/discovery/MulticastUtil.java
+++ b/agent/core/src/main/java/org/jolokia/discovery/MulticastUtil.java
@@ -5,6 +5,7 @@ import java.net.*;
 import java.util.*;
 import java.util.concurrent.*;
 
+import org.jolokia.config.ConfigKey;
 import org.jolokia.util.LogHandler;
 import org.jolokia.util.NetworkUtil;
 import org.jolokia.util.QuietLogHandler;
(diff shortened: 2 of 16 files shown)


agent/core/src/main/java/org/jolokia/config/ConfigKey.java
```diff
@@ -197,6 +197,16 @@
      */
     DISCOVERY_AGENT_URL("discoveryAgentUrl",true,false),
 
+    /**
+     * IPv4 Address for Jolokia's Multicast group.
+     */
+    MULTICAST_GROUP("multicastGroup",true,false),
```
rhuss:
```suggestion
    MULTICAST_GROUP("multicastGroup",true,false, "239.192.48.84"),
```

You can directly add a default value to the overloaded enum constructor with 4 args.

agent/core/src/main/java/org/jolokia/config/ConfigKey.java
```diff
@@ -197,6 +197,16 @@
...
+
+    /**
+     * Multicast port where to listen for queries.
+     */
+    MULTICAST_PORT("multicastPort",true,false),
```
rhuss:
```suggestion
    MULTICAST_PORT("multicastPort",true,false,"24884"),
```

agent/core/src/main/java/org/jolokia/discovery/MulticastUtil.java
```diff
@@ -17,22 +18,40 @@
...
-    public static final int JOLOKIA_MULTICAST_PORT = 24884;
+    // Default multicast port where to listen for queries
+    public static final int DEFAULT_JOLOKIA_MULTICAST_PORT = 24884;
 
     // Utility class
     private MulticastUtil() {
     }
 
+    private static <T> T getOrDefault(ConfigKey configKey, String sysEnvKey, T defaultValue) {
+        String property = System.getProperty("jolokia." + configKey.getKeyValue());
```
rhuss: This lookup is already performed upwards in the stack and shouldn't happen here.

agent/core/src/main/java/org/jolokia/discovery/MulticastUtil.java
```diff
@@ -17,22 +18,40 @@
...
+               return (T) Integer.valueOf(property);
+            } else {
+                return (T) property;
+            }
+        }
+        return defaultValue;
+    }
+
     static MulticastSocket newMulticastSocket(InetAddress pAddress, LogHandler pLogHandler) throws IOException {
+        String multicastGroup = getOrDefault(ConfigKey.MULTICAST_GROUP, "JOLOKIA_MULTICAST_GROUP", DEFAULT_JOLOKIA_MULTICAST_GROUP);
```
rhuss: This utility class is not the proper place for looking up configuration values. Instead this method should have two addition parameters, the group and the port. And feed in by the caller of this method.

rhuss: So the [method initializing the multicast thread](https://github.com/rhuss/jolokia/blob/ec1cebda618e68d12c9b1906820564408cf950e1/agent/core/src/main/java/org/jolokia/http/AgentServlet.java#L198) should pass through the `Configuration` object to the `MulticastSocketListenerThread` where it then be used to call this utility method.

I know its a bit tedious, but it helps in keeping the configuration handling separate.

rhuss:
btw, if you do a rebase to the latest Jolokia master, the CI run will be fixed.

graben:
I'm working on it but it will make the change more difficult and big.

graben:
@rhuss : Ready to merge? :-)

rhuss:
sorry, just back from a conference. will give it a review asap.

PR #436, opened 2020-04-27
Fixes SSL failure to connect using proxy or .war

### LINKED ISSUE #109: Add SSL support for the JSR-160 proxy
It would be nice, if the the proxy could communicate via SSL with the JSR-160 backend. 

For this to work, the environment for the connector must probably be extended (see [here](https://github.com/rhuss/jolokia/blob/master/agent/jsr160/src/main/java/org/jolokia/jsr160/Jsr160RequestDispatcher.java#L115) as entry point) and there must be a way to specify the path to the keystore holding the client's private key. It is probably the easiest to start by putting all these parameters in the request as well in order to let the proxy be as much agnostic as possible, not requiring any front up configuration. 

If SSL is to be used, then someone must put the keystore to a place where it is filesystem-reachable from within the proxy.

This issue has low priority, but I'm accepting pull requests (with unit tests) for this feature.

jc-roman:
Adds SSL-RMI support when the `javax.net.ssl.trustStore` system property is defined and fixes error "java.rmi.ConnectIOException: non-JRMP server at remote endpoint. Tested with tomcat 9 by mounting build .war file.  

Resolves #109 - Add SSL support for the JSR-160 proxy

### DIFF
--- a/agent/jsr160/src/main/java/org/jolokia/jsr160/Jsr160RequestDispatcher.java
+++ b/agent/jsr160/src/main/java/org/jolokia/jsr160/Jsr160RequestDispatcher.java
@@ -32,6 +32,7 @@ import java.util.regex.Pattern;
 import javax.management.*;
 import javax.management.remote.*;
 import javax.naming.Context;
+import javax.rmi.ssl.SslRMIClientSocketFactory;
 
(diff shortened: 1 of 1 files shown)


rhuss:
Thanks, that makes sense to me.

Issue #439, opened 2020-06-18
Find Hostaddress for HTTP bind by NIC interfacename


G-Ork:
The actual configuration only supports setting of an fixed ip address. In cluster environments host could have more ip for separate networks. It would be a relaxation to define a nic name match regex-pattern to find the bind ip address. 

In some environments it isn't possible or desired to configure each JVM for each node individually. Using the name of the network interface is an solution to unify the configuration.

| Host   | IP: ethA | IP: ethB |
|--------|----------|--------------|
| Node-A | 1.0.0.1  | 2.0.0.1      |
| Node-B | 1.0.0.2  | 2.0.0.2      |

I'd like to config the jvm agent like: 
NicMatch=ethB so jolokia picks up the 

IP: 2.0.0.1 for Host=Node-A and 
IP: 2.0.0.2 for Host=Node-B.

https://docs.oracle.com/javase/tutorial/networking/nifs/listing.html

rhuss:
Sounds like a good feature to have. Unfortunately I don't have time to implement this, but would be happy to help to integrate a pull request.

G-Ork:
I'll put it on my list then.

G-Ork:
PR is delivered. I could test the new features for me but i have problems to build the whole reactor due to missing mule dependencies. 

As long you support JRE 1.6 you should add maven toolchain plugin. Maven needs 1.8. Its quite dangerous to use an 1.8 jdk to compile 1.6 code.

grgrzybek:
Good news - with #751 fixed, you can now simply use this when running agent:
```shell
-javaagent:/path/to/jolokia-agent.jar=host=\${ip:ethB}
```

Mind that escaping is needed if used in command line, so your shell doesn't try to resolve `ip:ethB` variable.

You can also use:
```shell
-javaagent:/path/to/jolokia-agent.jar=config=/path/to/config.properties
```

and have `config.properties like this:
```properties
host=${ip:ethB}
```

PR #442, opened 2020-06-21
#439 HTTP_bind_by_NIC_interfacename

G-Ork:
- Delegate address resolving to an global service
- Added service for RegEx matching on IPs
- Added service for RegEx matching on interface names
- Extend DocBook for jvm-agent config

### DIFF
--- a/src/docbkx/agents/jvm.xml
+++ b/src/docbkx/agents/jvm.xml
@@ -114,12 +114,37 @@ java -javaagent:agent.jar=port=7777,host=localhost]]></programlisting>
         <td><constant>host</constant></td>
         <td>
           Hostaddress to which the HTTP server should bind to. If "*" or "0.0.0.0" is
-          given, the servers binds to every network interface.
+          given, the servers binds to every network interface. This setting is mutually exclusive to <em>ipmatch</em> and <em>nicmatch</em>. 
+          Do not combine this settings. Bad configuration or errors resolving network informations will fallback to loopback interface for security reasons.
         </td>
         <td>
           <constant>localhost</constant>
         </td>
         </tr>
+        <tr>
+        <td><constant>ipmatch</constant></td>
+        <td>
+          Indirect configuration of Hostaddress to which the HTTP server should bind to. 
+          The IP address string in textual presentation of all available network interfaces will be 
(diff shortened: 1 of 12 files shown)


rhuss:
Thanks a lot for your contribution ! I have a busy week ahead but will reserve some time on Friday to work integrate the PR (and also think about what todo with Java 6 support)

have a nice week ...

src/docbkx/agents/jvm.xml
```diff
@@ -114,12 +114,37 @@ java -javaagent:agent.jar=port=7777,host=localhost]]></programlisting>
...
-          given, the servers binds to every network interface.
+          given, the servers binds to every network interface. This setting is mutually exclusive to <em>ipmatch</em> and <em>nicmatch</em>. 
+          Do not combine this settings. Bad configuration or errors resolving network informations will fallback to loopback interface for security reasons.
         </td>
         <td>
           <constant>localhost</constant>
         </td>
         </tr>
+        <tr>
+        <td><constant>ipmatch</constant></td>
```
rhuss, line 125: I wonder whether we could avoid to a new configuration key like `nicmatch` or `ipmatch` and fold everything into the existing `host` variable. This approach would have multiple advantages:

* Multiple config-clashes would be avoided automatically
* Don't blow up the already huge configuration surface

So my suggestion would be to introduce multiples prefix like `nic:` or `ip:` which then trigger the different modes. Maybe even better just a `match:` if from the value it could be determined whether its a mac or an IP (by the existence of a `:` in the value ?)

Wdyt ?

G-Ork:
Hi thanks for the review and your feedback. I've stubled over many of the topics you suggested during the implementation. 

I don't mind to consider your suggestion. It would be alot easier to reuse the existing key. A prefix would be the best workaround to extend the matching without conflict with the old style dots in ip matching and the RegEx dots in the alternative matching options. 

Did you see a need for supporting to classify ipv4 or ipv6? For myself I decided it for the whole JVM. So I don't need a config key to only bind ipv4.



I would extend the documentation by a table with examples. This should be clearly to understand. 

I try to change it next week. 
I'm not allowed to code this weekend 😍

tadayosi:

Issue #446, opened 2020-08-31
Issue with JmxAdapter Unit Tests on JDK 8 & 9


rhuss:
There is an issue with a unit tests that are not working on older JDK: https://travis-ci.org/github/rhuss/jolokia/jobs/719557630

@skarsaune do you have an idea what is going wrong here ?

skarsaune:
Yeah, couldn't find the comment on github, trying to reply by mail. Seemed
to build on last push. I have a script that runs the tests on jdk 7-14 on
my mac. Can try on my linux as well.

rhuss:
Thanks ! I also found out that the OpenJDK 8 in Travis is different than mine that i have locally. E.g. i get NPE in unit tests on Travis but not locally.

rhuss:
Maybe its worth to find out exactly the JDK version that travis is using ?

skarsaune:
I have not been able to recreate identical problems. However I have looped over all jdks on my linux box. And in newer versions it fails due to a NamingException instead of a CommunicationException

`export` JAVA_HOME=/usr/lib/jvm/jdk1.8.0_261
mvn test-compile > /dev/null
for jdk in $(ls /usr/lib/jvm)
do
    export JAVA_HOME=/usr/lib/jvm/$jdk
    if [ -d $JAVA_HOME -a -f $JAVA_HOME/bin/java ]
    then
	mvn test --fail-at-end > /tmp/$jdk 2>/dev/null
	if [ $? -eq 0 ]
	then
	    echo "$jdk: PASS"
	else
	    echo "$jdk: FAIL"
	    cat /tmp/$jdk | grep -A 1 'sec  <<< FAILURE!'
	fi
    fi
done`

OUTPUT: 
`java-1.11.0-openjdk-amd64: FAIL
simpleWhiteListWithConfig(org.jolokia.jsr160.Jsr160RequestDispatcherTest)  Time elapsed: 0.044 sec  <<< FAILURE!
java.lang.AssertionError: expected [true] but found [false]
--
simpleWhiteListWithSysProp(org.jolokia.jsr160.Jsr160RequestDispatcherTest)  Time elapsed: 0.004 sec  <<< FAILURE!
java.lang.AssertionError: expected [true] but found [false]
java-1.13.0-openjdk-amd64: FAIL
simpleWhiteListWithConfig(org.jolokia.jsr160.Jsr160RequestDispatcherTest)  Time elapsed: 0.032 sec  <<< FAILURE!
[...]

PR #450, opened 2020-10-17
Issue #446 : future proof some tests for newer openjdk distributions …

skarsaune:
…(different exception for timeout of jsr160 connectons)

While looking at #446 : was not able to reproduce the exact issues seen in travis CI, but brute force tested other linux openjdk versions and found one future problem.


rhuss (approved):
thanks ! 

lgtm

PR #456, opened 2021-04-28
Unblock building with Maven 3.8.1

ryandgoulding:
Bumps source/target versions of `maven-compiler-plugin` from `1.6` to `1.7`, as `1.6` is no longer supported.

### DIFF
--- a/agent/jvm/pom.xml
+++ b/agent/jvm/pom.xml
@@ -96,8 +96,8 @@
       <plugin>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <source>1.6</source>
-          <target>1.6</target>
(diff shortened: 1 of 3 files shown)


rhuss:
As mentioned elsewhere,  this bump might fix the issue at a hand but has more severe consequences for people relying still on Java 6 (and yes, this still happens). At least it needs to be added to the documentation, any potential tests with Java 6 should be removed, and the changelog updated. Also, this might justify a bump in the minor version.

I wonder whether you could live for the moment to run with a jdk that is still compatible with Java 6 ?

tadayosi:
This is superceded by #547. Let me close this one. Thanks.

PR #458, opened 2021-04-28
Ensure InputStreamReader is closed on method exit

ryandgoulding:
Uses try-with-resources to ensure the InputStreamReader is closed upon exiting the try-block.


ryandgoulding:
@rhuss Is JDK 6 still supported?  If so, we can handle this using an alternative method.

rhuss:
In Java 6 there is no try-resource statement so you probably better use the classical try-catch-finally style for now.

PR #460, opened 2021-04-28
Fix possible integer overflow in OperationChecking.java

ryandgoulding:
By using a long constant instead of an int, seconds will be
type-upgraded to a long as well.


rhuss (commented):
Thanks ! Looks good to me (and sorry for the delay)

PR #462, opened 2021-05-15
In case the kubernetes API url has no explicit port, do not attempt to map explicit port

skarsaune:


rhuss (approved):
Loogs good to me, thanks !

Issue #394, opened 2018-10-17
JDK11 reflection warnings


DagW:
We recently updated our application to run with AdoptOpenJDK 11,
upon start the following lines appeared

```
Oct 17 11:29:14 XXX-01 java[58654]: WARNING: An illegal reflective access operation has occurred
Oct 17 11:29:14 XXX-01 java[58654]: WARNING: Illegal reflective access by org.jolokia.jvmagent.client.command.AbstractBaseCommand (file:/usr/local/libexec/jolokia-jvm-agent.jar) to method sun.tools.attach.HotSpotVirtualMachine.getSystemProperties()
Oct 17 11:29:14 XXX-01 java[58654]: WARNING: Please consider reporting this to the maintainers of org.jolokia.jvmagent.client.command.AbstractBaseCommand
Oct 17 11:29:14 XXX-01 java[58654]: WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
Oct 17 11:29:14 XXX-01 java[58654]: WARNING: All illegal access operations will be denied in a future release
```

Thoughts?

bfg111:
Seeing this error as well
Java version:
```
java 11.0.1 2018-10-16 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.1+13-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.1+13-LTS, mixed mode)
```

Jolokia version:
```
Jolokia JVM Agent 1.6.0
```

```
Mar 22 14:54:50 xxxxxxxx java[26349]: WARNING: An illegal reflective access operation has occurred
Mar 22 14:54:50 xxxxxxxx java[26349]: WARNING: Illegal reflective access by org.jolokia.jvmagent.client.command.AbstractBaseCommand (file:/[base_location]/jolokia-jvm-1.6.0-agent.jar) to method sun.tools.attach.HotSpotVirtualMachine.getSystemProperties()
Mar 22 14:54:50 xxxxxxxx java[26349]: WARNING: Please consider reporting this to the maintainers of org.jolokia.jvmagent.client.command.AbstractBaseCommand
Mar 22 14:54:50 xxxxxxxx java[26349]: WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
Mar 22 14:54:50 xxxxxxxx java[26349]: WARNING: All illegal access operations will be denied in a future release
[...]

PR #464, opened 2021-06-24
[#394] Provide non-reflective access to JVM Attach API on JDK9+

grgrzybek:


grgrzybek:
Please do not merge yet. Work in progress.

agent/jvm/pom.xml
```diff
@@ -96,16 +96,15 @@
       <plugin>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <source>1.6</source>
-          <target>1.6</target>
+          <source>1.7</source>
+          <target>1.7</target>
```
rhuss, line 100: I guess this jump is required, and I'm not against it. But we need to bump the version number accordingly, to indicate that the release is not compatible with Java 1.6. anymore. Also, I'd prefer to have this as a global switch for all components (not only for the JVM agent)

agent/jvm/src/main/java/org/jolokia/jvmagent/security/asn1/DERInteger.java
```diff
@@ -0,0 +1,99 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jolokia.jvmagent.security.asn1;
+
+public class DERInteger implements DERObject {
```
rhuss, line 19: Are those ASN1 objects copied over from some other open-source library (like bouncycastle) or do you have written those ? 

It's probably over the top here, but are there any tests for those classes ?

rhuss (commented):
Thanks a lot for all of your efforts, looks good to me in general ! Maybe it would have been better to split it up in two PRs (one for the reflection in the VirtualMachineHandler, one for the CERT handling).

I have two questions though, see the comments inline. Also if the ASN.1 objects are borrowed from somewhere else we should check the license and add the corresponding license headers/files. For the switch to 1.7 I wonder whether this is required, and if so, we should do it at the top-level.

agent/jvm/pom.xml
grgrzybek: It is required, when I tried to run the tests using JDK17 (as it no longer can handle `target=1.6`), but it's not a requirement for any other change. I can skip this change.

agent/jvm/src/main/java/org/jolokia/jvmagent/security/asn1/DERInteger.java
grgrzybek: The tests are in `org/jolokia/jvmagent/security/asn1/DEREncodingTest.java` and I confirm - I looked a bit at BC classes (only to check what _DER Tag_ is), but every line of code is written by me. I was consulting only the X.680/X.690 specifications.

grgrzybek:
First - this PR can be treated as "for review only" and I agree that separate PRs are better.

DER classes are written by me, as I didn't want to borrow anything from BC (or internal JDK classes). I also didn't want to introduce any dependency to external, lightweight ASN.1/DER libraries.
There are few missing DER/ASN.1 constructs, because I needed only the ones for X.509 certificate structure.

grgrzybek:
I'm just checking the final part - `@JsonMBean` annotation handling. When I'm ready, I'll update this PR (force pushing probably), but eventually I'll split this PR into 2 (or 3) separate ones.

rhuss:
> DER classes are written by me, as I didn't want to borrow anything from BC (or internal JDK classes). I also didn't want to introduce any dependency to external, lightweight ASN.1/DER libraries.

ah, very cool. Do you think we should some unit tests for those ? They don't look trivial

rhuss:
I think we can keep it 1 PR, don't bother with fiddling around with multiple PRs. Maybe you can have several commits per feature, but happy also if not. Don't waste efforts here.

grgrzybek:
> ah, very cool. Do you think we should some unit tests for those ? They don't look trivial

the tests are in `org/jolokia/jvmagent/security/asn1/DEREncodingTest` ;)

grgrzybek:
And #466 is a separate PR for cert generation.

agent/jvm/pom.xml
rhuss: Maybe we can skip it for now, and consider to bump to 1.7 as minimum in a followup PR (and then already on the overall level) ?

grgrzybek: I reverted this change, so it's still 1.6. 1.7 was needed if I wanted to build using JDK17. But I needed it to _test_ using JDK17. After verifying it works, we can leave at 1.6 value ;)

rhuss:
Sorry that I overlooked the tests ;( going to continue tomorrow on the PRs ...

rhuss (approved):
Looks good to me !

PR #466, opened 2021-07-05
[#394] Reimplement self-signed certificate generation without using r…

grgrzybek:
…eflection


rhuss (approved):
Thanks a ton !

grgrzybek:
you're welcome! I've had an opportunity to learn how to DER-encode an OID identifier ;)

PR #489, opened 2021-12-05
AKS: No longer rely on selflink of pod

### LINKED ISSUE #486: Pod metadata selflink stopped working in Azure AkS and GKE
Solution: Manually create URL instead of relying on selfLink

skarsaune:
# Changes

- :bug: Handle changes in AKS and GKE

/kind bug

Fixes #486

**Release Note**

```release-note

Fixes bug that made it impossible to connect to pods in later versions of AKS and GKE
```

### DIFF
--- a/client/kubernetes/src/main/java/org/jolokia/kubernetes/client/KubernetesJmxConnector.java
+++ b/client/kubernetes/src/main/java/org/jolokia/kubernetes/client/KubernetesJmxConnector.java
@@ -1,5 +1,6 @@
 package org.jolokia.kubernetes.client;
 
+import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.client.BaseClient;
 import io.fabric8.kubernetes.client.Config;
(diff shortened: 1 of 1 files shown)


client/kubernetes/src/main/java/org/jolokia/kubernetes/client/KubernetesJmxConnector.java
```diff
@@ -118,15 +119,13 @@ protected J4pClient expandAndProbeUrl(KubernetesClient client,
...
+  public static StringBuilder buildProxyPath(Pod pod, String protocol, String port, String path) {
+    final ObjectMeta metadata = pod.getMetadata();
+    final StringBuilder url = new StringBuilder("/api/").append(pod.getApiVersion()).append("/namespaces/").append(metadata.getNamespace()).append("/pods/");
     if (protocol != null && !protocol.equals("http:")) {
-      //if we need to specify protocol, inject in url
-      final int index = url.lastIndexOf(pod.getMetadata().getName());
-      if (index > -1) {
-        url.insert(index, protocol);
-      }
+      url.append(protocol);
```
rhuss, line 126: Haven't seen the protocol in API URLs really before. Why do you need it here ?

skarsaune: It allows you to specify https instead of http

skarsaune:
The self link is not part of the API even in docker-desktop now, so even more urgent now

rhuss:
thanks!

PR #492, opened 2022-01-10
Fix error for invalid executor in config

a1exsh:
# Changes

- :broom: Update error message: fix typo and make it look more like the other messages in the same file.

/kind cleanup


rhuss (approved):
Thanks!

PR #502, opened 2022-04-23
Enable MAVEN Parallel Test

shisheng-1:
According to [Maven parallel test](https://www.baeldung.com/maven-junit-parallel-tests), we can run tests in parallel.


rhuss:
Looks good to me, but please could you rebase. The tests should be fixed then.

PR #543, opened 2023-02-06
Improve proxy denylist

qtc-de:
### Changes

Jolokia attempts to prevent LDAP based JNDI injection attacks by applying a default denylist to JMX service URLs that are supplied in proxy-mode. However, the pattern matching that is done with this list is case sensitive. This allows for an easy bypass of the protection. The following JMX service URL passes the filter and causes an outbound LDAP connection:

```
service:jmx:Rmi:///jndi/ldap://bypassed/ups
```

I hope you don't mind the public disclosure, but I estimate the impact of this issue quite low.

/kind enhancement

### Release Note

```release-note
The white and blacklist that user supplied JMX service URLs are matched against when running Jolokia in proxy mode are now treated case insensitive. This could cause different behavior for already existing configurations. If you use Jolokia in proxy mode and rely on specific white or blacklist settings, you should review your configuration.
```


rhuss (approved):
Good catch, thanks!

Issue #524, opened 2022-11-02
Support JDK 17


apupier:
JDK 17 is the latest LTS (Lon-term Support) released as of today.
it would be nice to support it

rhuss:
Definitely a +1 from my side.

varikin:
I am rolling out JDK 17 and tested our current version, Jolokia 1.6.2. Basic functionality is working of getting JMX over HTTP as JSON. Is there more that I should be testing related to 17 and Jolokia?

grgrzybek:
Tests work on JDK 1.7:
```
[INFO] Reactor Summary:
[INFO] 
[INFO] jolokia-parent ..................................... SUCCESS [  0.191 s]
[INFO] jolokia-test-util .................................. SUCCESS [  1.211 s]
[INFO] jolokia-agent-parent ............................... SUCCESS [  0.005 s]
[INFO] jolokia-core ....................................... SUCCESS [ 20.601 s]
[INFO] jolokia-jsr160 ..................................... SUCCESS [  1.192 s]
[INFO] jolokia-war ........................................ SUCCESS [  0.390 s]
[INFO] jolokia-war-unsecured .............................. SUCCESS [  0.055 s]
[INFO] jolokia-osgi ....................................... SUCCESS [  3.577 s]
[INFO] jolokia-osgi-bundle ................................ SUCCESS [  1.293 s]
[INFO] jolokia-jmx ........................................ SUCCESS [  2.142 s]
[INFO] jolokia-jvm ........................................ SUCCESS [ 28.112 s]
[INFO] jolokia-spring ..................................... SUCCESS [  6.042 s]
[INFO] jolokia-mule ....................................... SUCCESS [  2.434 s]
[INFO] jolokia-it-parent .................................. SUCCESS [  0.004 s]
[INFO] jolokia-it-core .................................... SUCCESS [  0.345 s]
[INFO] jolokia-it-war ..................................... SUCCESS [  0.076 s]
[INFO] jolokia-client-parent .............................. SUCCESS [  0.004 s]
[INFO] jolokia-client-java ................................ SUCCESS [  8.701 s]
[INFO] jolokia-client-javascript .......................... SUCCESS [  0.349 s]
[INFO] jolokia-jmx-adapter ................................ SUCCESS [ 38.016 s]
[INFO] jolokia-kubernetes ................................. SUCCESS [  6.582 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[...]

PR #547, opened 2023-06-29
Ggrzybek build,test improvements

grgrzybek:


grgrzybek:
@rhuss @tadayosi FYI - I did some review that we indeed run on JDK 1.6 and later. See #524.

From here we can consciously drop old JDK support if desired.

client/jmx-adapter/src/test/java/org/jolokia/client/jmxadapter/JmxBridgeTest.java
```diff
@@ -242,19 +242,44 @@ public static Object[][] platformMBeans() {
...
 
 
   @DataProvider
   public static Object[][] safeOperationsToCall() {
+    String version = System.getProperty("java.specification.version");
+    if (version.contains(".")) {
+      version = version.substring(version.lastIndexOf('.') + 1);
+    }
+    int v = Integer.parseInt(version);
+    if (v <= 6) {
```
tadayosi, line 263: This trick seems to be used in different places, so in the future it might be worth moving to a test helper class so other tests can share the code.

tadayosi (approved):
Great!

client/jmx-adapter/src/test/java/org/jolokia/client/jmxadapter/JmxBridgeTest.java
grgrzybek: ah yes - I decided to do it when there's a need to use it in 3rd place ;)

tadayosi: good old criteria -- third time you repeat yourself, abstract it to shared code ;-)

PR #561, opened 2023-07-17
Fix of typo in changes.xml

rhuss:


rhuss:
Merging because trivial fix.

Issue #567, opened 2023-08-21
Split JavaScript client from the project and into a dedicated npm project


tadayosi:
Right now we have the JS client library under the same project with the Java agents and clients:
https://github.com/jolokia/jolokia/tree/main/client/javascript

But since the JS client should have a different release/update cadence than the Java artifacts and also JS has a different tooling ecosystem from Java's such as nodejs, npm, yarn, and testing frameworks like jest/karma/jasmine/cypress, IMO it's easier to make it a dedicated github project as a npm package and release it through the public npm registry. I think V2 is the best time if we go the path.

Ideas to work on if we split the JS client to a different project:

- Modernise the project by converting it to TypeScript
- Rewrite tests with popular testing frameworks such as Jest and Cypress
- Release it through `npm publish` to https://www.npmjs.com/ so that the client JS app can use it just by `npm i jolokia`
- Stop depending on jQuery and use the builtin Fetch API https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API so that the client app doesn't need to bother depending on jQuery for just using the Jolokia JS client

rhuss:
Now that we have moved over to our own GitHub organisation I think that makes totally sense, not only for the JavaScript client library but also for other client libraries like Perl (jmx4perl), Python (e.g. https://github.com/wbrefvem/python-jolokia) or even the Java client library (but this needs to be discussed whether it wouldn't be better to keep the Java client library in the main repo).

I would also love to settle on a naming convention for the repository names for client libraries, like

* `jolokia/jolokia-client-javascript` (to reflect jolokia also in the checked-out directory name)
* `jolokia/client-javascript` (to have a shorter name, however the name only makes sense in the context of the GitHub orga then)

I think its also okay to switch the build tooling to `npm`, however, I would also love to publish the client as a webjar (https://www.webjars.org/) to make it easily embeddable in Java web backends.. Though I'm not sure if webjars are still a thing (i.e are actively maintained and propagated)

tadayosi:
Thanks @rhuss for your comment! +1 for moving other client libraries as well, especially jmx4perl is a good candidate to move to the hawtio org if you'd like. For the Java client, I don't have a strong opinion about moving it. For the time being, we can perhaps keep it in the same Java project with the agents but I'm not also against moving it to a separate project all at once.

[...]

PR #574, opened 2023-09-22
Define .d.ts for JS client #567

tadayosi:
Initial attempt to provide `.d.ts` for `jolokia.js` and `jolokia-simple.js`. Since Cubism.js doesn't seemt to provide type definitions itself yet, it shouldn't make much sense to provide the type defitions for `jolokia-cubism.js` as well, so it should be acceptable not to provide it for `jolokia-cubism.js` as the first step.

Note:
- Type defitions for `jolokia-simple.js` is merged into a single `index.d.ts` as otherwise the definitions would become a bit more complicated to define.

Showcase release:
- https://www.npmjs.com/package/jolokia.js/v/2.0.0-dev.4

Relates #567

### DIFF
--- a/client/javascript/package-lock.json
+++ b/client/javascript/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "jolokia.js",
-  "version": "2.0.0-dev.5",
+  "version": "2.0.0-dev.6",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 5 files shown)


grgrzybek (approved):
wow!

grgrzybek:
Test failures are my fault - soon to be fixed.
Thanks @tadayosi !

Issue #575, opened 2023-09-25
Remove usages of jQuery from JS client as much as possible


tadayosi:
The JavaScript client should stop depending on jQuery and use the builtin Fetch API https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API, so that the client app doesn't need to bother depending on jQuery for just using the Jolokia JS client.

grgrzybek:
@tadayosi please add very quick info about such changes to https://github.com/jolokia/jolokia/blob/2.0/src/site/asciidoc/migration.adoc - at some point (soon) I'll prepare proper "changes in 2.0" page.

tadayosi:
What remain to be done:

- [ ] `$.extend`
- [ ] `$.ajax`

tadayosi:
The current `jolokia.js` provides both sync and async modes, which is based on the `$.ajax` feature. I found it hard to completely replace it with Fetch API as it only supports async API. So forcing replacing `$.ajax` with Fetch API will bring destructive, backward incompatible changes to `jolokia.js`, which we should avoid.

It should be still preferable to have removed the rest of jQuery API usages from the JS client to minimise dependency to external libs.

Instead, I propose to create a yet another base JS client `jolokia-fetch.js` which is based on Fetch API and only supports async mode. Users who don't want to bring dependency to jQuery for their apps will be able to use this version of base JS client. `jolokia-simple.js` should be compatible with both versions, now that it itself doesn't have dependency to jquery.

@grgrzybek @rhuss FYI

grgrzybek:
:+1:

tadayosi:
I'll close this and follow it up with #580.

PR #578, opened 2023-09-27
Removing usages of jQuery from jolokia.js and jolokia-simple.js

tadayosi:
Relates #575. Cleaning up as many usages of jQuery functions as possible from those JS scripts.

What remain to be done are:

* `$.extend`
* `$.ajax`

image


grgrzybek (approved):
wow - that's really cool!

tadayosi:
@grgrzybek thanks!

PR #591, opened 2023-10-03
Resolve warnings and apply suggestsions from IDEA based on Java 11

### LINKED ISSUE #592: Resolve warnings and apply suggestsions from IDEA based on Java 11
I found that there are a bunch of warnings and suggestions from IDEA on the 2.0 codebase. Since we are now based on Java 11, I think many of them can be applied and it's good time to do so.

Most of them should be just automatic applications of IDEA's suggestions. I expect they are identical in terms of functionality and thus safe to apply.

tadayosi:
I found that there are a bunch of warnings and suggestions from IDEA on the 2.0 codebase. Since we are now based on Java 11, I think many of them can be applied and it's good time to do so. I'm going ahead and applying them to the codebase.

Most of them are just automatic applications of IDEA's suggestions. I expect they are identical in terms of functionality and thus safe to apply.

* Collaterally, it also fixes a bug with NotificationDataUpdater where LIST request didn't collect notification info in the response.

Fix #592

### DIFF
--- a/agent/jvm/src/main/java/org/jolokia/jvmagent/CleanupThread.java
+++ b/agent/jvm/src/main/java/org/jolokia/jvmagent/CleanupThread.java
@@ -29,12 +29,12 @@ import com.sun.net.httpserver.HttpServer;
  */
 class CleanupThread extends Thread {
 
-    private HttpServer server;
-    private ThreadGroup threadGroup;
+    private final HttpServer server;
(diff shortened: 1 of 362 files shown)


grgrzybek:
Big :+1: for such change - I always preferred dedicated issue/commit(s) for such changes and I never liked reformatting entire code for one line fix ;)

PR #597, opened 2023-10-10
Fix typo Javascript --> JavaScript

apupier:


examples/client-javascript-test-app/src/main/javascript/support/qunit.js
```diff
@@ -2056,7 +2056,7 @@ function inArray( elem, array ) {
 }
 
 /*
- * Javascript Diff Algorithm
+ * JavaScript Diff Algorithm
```
rhuss, line 2059: Same here. Just for license sanity we might not want to touch dependent files (but well, probably a bit nit picky)

rhuss (approved):
Looks with some little nit pick (you can ignore that though, guess its not super important)

PR #656, opened 2024-02-07
Request with HTTPS Origin over HTTP should be rejected for security

tadayosi:
CORS best practices suggest that if Jolokia receives requests over HTTP that have `Origin: https://...` it should reject them, because Origin=https implies they come from a secure website but receiving the requests over HTTP from there would result in compromising confidentiality and integrity over HTTPS.

I implemented the protection at `HttpRequestHandler#checkAccess()` directly as it is the only location where both the request scheme and origin header value are available. Ideally, `PolicyRestrictor`'s `CorsChecker` should have such checking but the `Restrictor` interface only accepts origin values at this moment, and implementing the checking at `CorsChecker` causes changes and incompatibility to the public interface.

We might want to move the checking logic to `CorsChecker` in the future to make it optional, but for now I think it's fine to have it at `HttpRequestHandler`.


tadayosi:
Please review this @grgrzybek @rhuss @apupier

grgrzybek (approved):

grgrzybek:
I'm sorry @tadayosi - I'll do 2.0.2 this week (at latest - during weekend).

PR #674, opened 2024-04-04
Make KeyStoreUtil.updateWithServerPems load the full cert chain

shayelkin:
updateWithServerPems uses CertificateFactory.GenerateCertificate(), which only loads the first certificate in the server certificate file. When using a certificate that is signed by an intermediate CA (like Let's Encrypt,) this would result in a certifiacte that won't be accepted by clients that only has the root CA, and not the intermediate CA, in their trusted certificates stores.

This replaces that call with a call to CertificateFactory.GenerateCertificates(), that loads the full chain, allowing clients with just the root chain to trust the server.


tadayosi (approved):
LGTM.

PR #685, opened 2024-05-31
docs: Fix option name: canonicalProperties -> canonicalNaming

tadayosi:
@grgrzybek Just please check if the fix is correct.


grgrzybek:
Good catch, thanks!
https://github.com/jolokia/jolokia/blob/v2.0.2/server/core/src/main/java/org/jolokia/server/core/config/ConfigKey.java#L142

grgrzybek (approved):

Issue #686, opened 2024-05-31
Replace json-simple with different JSON library


tadayosi:
I noticed that Jolokia is still based on the old json-simple library for JSON manipulations. json-simple is no longer maintained, of which the last release was more than ten years ago. We should migrate to a newer and maintained JSON library such as [JSON-java](https://github.com/stleary/JSON-java). JSON-java seems to have quite similar API to json-simple and appears mostly compatible, so migration shouldn't be so hard.

There are outstanding issues with json-simple as well, for instance:
- https://github.com/jolokia/jolokia/issues/226

Hopefully they might be resolved by migrating to JSON-java as well.

grgrzybek:
:+1:
we definitely need something simple - no data binding, no execution, just json - hashmap.
I'll check next week (PTO today).

tadayosi:
One more motivation to move away from json-simple to JSON-java is that, while json-simple cannot parse JSON with native BigInt notation (it throws a parser exception):
```
{ "bigint": 12345678901234567890n }
```
JSON-java can parse it without problems (it parses the value to string "12345678901234567890n").

rhuss:
Happy to migrate, one of the reasons I've chosen json-simple was because of its size, since I wanted the agent to be as small as possible (below 100k was once a goal, now probably more like below 600k). I would still keep size as a criterion when selecting a new library

tadayosi:
Good to know that size is an important factor for the agent. Comparing json-simple with JSON-java:
- json-simple: ~24k
- JSON-java: ~78k

https://repo1.maven.org/maven2/com/googlecode/json-simple/json-simple/1.1.1/
https://repo1.maven.org/maven2/org/json/json/20240303/

So switching to JSON-java would increase the agent size by ~54k. We might want to examine all the dependencies once to see if there's any room for shrinking the size.

davsclaus:
In camel-core we have jsooner - it has no dependency at all
https://github.com/apache/camel/tree/main/tooling/camel-util-json

tadayosi:
Oh, `camel-util-json` is only ~40k. It's smaller than JSON-java.
https://repo1.maven.org/maven2/org/apache/camel/camel-util-json/4.6.0/

grgrzybek:
Hmm, no dependencies, but leads to Camel parent which may bring `<dependencyManagement>`...

davsclaus:
You can copy the code over and build it as part of hawtio

tadayosi:
[...]

PR #693, opened 2024-06-04
Add config option "serializeLong" to support accurate long values in JSON #495

### LINKED ISSUE #495: Jolokia JS client does not support Long/BigInt
Hi, there is an issue with JS client library when the Java agent returns a numeric value greater than [Number#MAX_VALUE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_VALUE)

This affects JMX attributes/operations when the server returns a LONG value for instance.

The issues stems from [jolokia.js#L226 ](https://github.com/rhuss/jolokia/blob/c963d57751ef29ad924844ec6319fddabd8c2085/client/javascript/src/main/javascript/jolokia.js#L226) where the default JSON parser is used.

To preserve the value, would it be possible to use a custom parser to parse numeric values into BigInt (there are libs for this), or convert numeric value to quoted String?

See related issue at: https://github.com/hawtio/hawtio-integration/issues/108

tadayosi:
Fix #495

This pull req is still a draft but let me share this early so that it could be checked by the experts.

Here are the highlights:

- I proposed `serializeLong` option with three values `none`, `string` and `bigint` at #495, but since the current json library json-simple doesn't support bigint notation in JSON (#686) I dropped `bigint` from this implementation. Now it only supports `none` and `string`.
- We may consider supporting `bigint` option in the future but we would also need to be aware that the bigint notation doesn't comply with [the standard JSON specification](https://www.json.org/json-en.html) and JSON is now effectively an universal message format that's not monopolized by JavaScript.

I'm working to add a few more commits to it:
- [x] Add client side option for `jolokia.js` that allows users to make the client return a string response instead of the `JSON.parse`'d object to provide yet another solution to the problem at the client side
- [x] Update docs with the newly added options

### DIFF
--- a/server/core/src/main/java/org/jolokia/server/core/service/serializer/SerializeOptions.java
+++ b/server/core/src/main/java/org/jolokia/server/core/service/serializer/SerializeOptions.java
@@ -47,17 +47,21 @@ public final class SerializeOptions {
     // Maximum number of objects to return
     private final int maxObjects;
 
+    // How to serialize long values: "none" or "string"
+    private final String serializeLong;
+
     // Handler which determines what should be done when
     // extracting of a value fails
     private final ValueFaultHandler faultHandler;
 
     // Use a builder to construct this object
-    private SerializeOptions(int pMaxDepth, int pMaxCollectionSize, int pMaxObjects,
+    private SerializeOptions(int pMaxDepth, int pMaxCollectionSize, int pMaxObjects, String pSerializeLong,
                              ValueFaultHandler pFaultHandler) {
         maxDepth = pMaxDepth;
         maxCollectionSize = pMaxCollectionSize;
         maxObjects = pMaxObjects;
         faultHandler = pFaultHandler;
+        serializeLong = pSerializeLong;
     }
 
     /**
@@ -92,6 +96,15 @@ public final class SerializeOptions {
     }
 
     /**
+     * Get the option for serializing long values.
+     *
+     * @return the option for serializing long values
+     */
+    public String getSerializeLong() {
+        return serializeLong;
+    }
+
+    /**
      * Get the configure fault handler which determines, how extractions fault are dealt with
      *
      * @return the configured fault handler
@@ -114,6 +127,8 @@ public final class SerializeOptions {
         private int maxDepth;
         private int maxCollectionSize;
         private int maxObjects;
+
+        private String serializeLong;
 
         private ValueFaultHandler faultHandler;
         private boolean useAttributeFilter;
@@ -138,6 +153,7 @@ public final class SerializeOptions {
             hardMaxDepth = pHardMaxDepth;
             hardMaxCollectionSize = pHardMaxCollectionSize;
             hardMaxObjects = pHardMaxObjects;
+            serializeLong = "none";
             faultHandler = ValueFaultHandler.THROWING_VALUE_FAULT_HANDLER;
         }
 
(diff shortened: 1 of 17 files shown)


grgrzybek:
Tomorrow I have to prepare urgent patch, but I'll review soon!

service/serializer/src/test/java/org/jolokia/service/serializer/json/ObjectToJsonConverterTest.java
```diff
@@ -186,6 +186,16 @@ public void convertTransientValue() throws AttributeNotFoundException {
...
     }
 
+    @Test
+    public void convertLong() throws AttributeNotFoundException {
+        long value = 900719925474099123L;
+        Object ret1 = converter.serialize(value, null, SerializeOptions.DEFAULT);
+        assertEquals(900719925474099123L, ret1);
+        SerializeOptions.Builder builder = new SerializeOptions.Builder();
+        Object ret2 = converter.serialize(value, null, builder.serializeLong("string").build());
+        assertEquals("900719925474099123", ret2);
```
grgrzybek, line 196: So at server side there's no place where `n` suffix is added?

tadayosi: Yes. At least so long as we use json-simple, if we insert `12345n` in JSON the last step of ejecting the response to the client should fail throwing a parser exception at the server side.

Also, even if we use JSON-java that supports the bigint notation, we would then no longer align with the JSON standard as there's no mention of the bigint notation for number in its original spec.

grgrzybek: So please remind me - currently jquery with `$.ajax` + `ajaxParams.dataType = 'json'` simply does `JSON.parse()` and there's no way to handle `n` suffix at all?
So when long is serialized into string, we should simply rely on type coercion when someone does something like `response.value.committed + 0`?

tadayosi: Yes, you are right. When we set the new option `serializeLong=string` then all long values are serialized into string, and it's all the responsibility of the client side to deserialize them into long or number with the user's prefered method.

But I'd say `serializeLong=string` will be never used with clients other than JS, as they don't have the issue with long values unlike JS.

tadayosi:
> Add client side option for jolokia.js that allows users to make the client return a string response instead of the `JSON.parse`'d object to provide yet another solution to the problem at the client side

Added support for returning the response as string to jolokia.js by unlocking `ajaxParams.dataType = 'text'` to the users. Then the JS client returns responseText instead of parsed JSON objects. It's intended to give users additional way to get along with the unsafe number issue with JS.

tadayosi:
Docs updated with minor fixes. Now ready for review and merge.

server/core/src/main/java/org/jolokia/server/core/service/serializer/SerializeOptions.java
```diff
@@ -47,17 +47,21 @@ public final class SerializeOptions {
     // Maximum number of objects to return
     private final int maxObjects;
 
+    // How to serialize long values: "none" or "string"
```
grgrzybek, line 50: Hmm, I'm just wondering if it should be `"number" or "string"`... `"none"` may be confusing. But it's not a big deal.

src/documentation/manual/modules/ROOT/pages/agents/jvm.adoc
```diff
@@ -327,6 +327,35 @@ Multiple cipher suites can be configured by using additional options with consec
 suffixes like in `sslCipherSuite.1`, `sslCipherSuite.2`, ...
 |
 
+|`maxDepth`
```
grgrzybek, line 330: were these 3 really missing? My fault ;) Thanks for checking!

server/core/src/main/java/org/jolokia/server/core/service/serializer/SerializeOptions.java
tadayosi: You are right. `number` should be better. I'll update it. Thanks!

grgrzybek (approved):

tadayosi:
Thanks, let me merge this.

PR #716, opened 2024-07-19
[Fixes #212] Support multi-dimensional arrays

### LINKED ISSUE #212: Unable to deserialize multi-dimension arrays
The StringToOpenTypeConverter doesn't seem to be able to handle arrays of more than one dimension, despite the fact that the ObjectToJsonConverter handles it just fine.

For example, an Integer[] works OK, but an Integer[][] will throw a java.lang.NumberFormatException when parsed.

The issue seems to lie in the org.jolokia.converter.object.ArrayTypeConverter.convertToObject method, which assumes the array only has one dimension and that the elements are of elementOpenType, which is wrong for arrays of dimension > 1.

It should check type.getDimension(), and handle the case where the dimension is > 1.

Minimal code to reproduce:

``` java
    StringToObjectConverter stoc = new StringToObjectConverter();
    ObjectToJsonConverter otjc = new ObjectToJsonConverter(stoc);
    StringToOpenTypeConverter stotc = new StringToOpenTypeConverter(stoc);
    Integer[] oneDim = new Integer[] { 1, 2 };
    // This access-restricted class is used to obtain the OpenType for arrays.
    final com.sun.jmx.mbeanserver.MXBeanMappingFactory f = com.sun.jmx.mbeanserver.MXBeanMappingFactory.DEFAULT;
    { // One dimension array, conversion works both ways
        System.out.println("one dim: " + Arrays.deepToString(oneDim));
        Object res = otjc.convertToJson(oneDim, null, JsonConvertOptions.DEFAULT);
        System.out.println("one dim to json: " + res);
        OpenType<?> type = f.mappingForType(oneDim.getClass(), f).getOpenType();
        System.out.println("one dim from json: " + Arrays.deepToString((Object[]) stotc.convertToObject(type, res)));
    }
    Integer[][] twoDims = new Integer[][] { oneDim, { 3, 4 } };
    { // Two dimension array, conversion only works one way
        System.out.println("two dims: " + Arrays.deepToString(twoDims));
        Object res = otjc.convertToJson(twoDims, null, JsonConvertOptions.DEFAULT);
        System.out.println("two dims to json: " + res);
        OpenType<?> type = f.mappingForType(twoDims.getClass(), f).getOpenType();
        System.out.println("two dims from json:" + Arrays.deepToString((Object[]) 
// This will throw a NumberFormatException
            stotc.convertToObject(type, res)));
    }
```

The StringToOpenTypeConverter.convertToObject() will throw the following exception for the two dimension array:
java.lang.NumberFormatException: For input string: "[1,2]"

grgrzybek:


tadayosi (approved):
Does it also support arrays with more than 3D?

rhuss:
Thanks! Looks good to me
/approve
/lgtm

grgrzybek:
> Does it also support arrays with more than 3D?

I hope so ;) I didn't check - the algorithm is recursive.

grgrzybek:
Yep. This works too:
```java
@Test
public void arrayType4D() throws Exception {
    // 2 elements
    int[] oneDim = new int[]{1, 2};
    // 3 x 2 elements
    int[][] twoDims = new int[][] { oneDim, { 3, 4 }, { 5, 6 } };
    // 2 x (3 x 2 elements) elements
    int[][][] threeDims = new int[][][] { twoDims, {{ 3, 4 }, { 5, 6 }, { 7, 8 } } };
    // 3 x (2 x (3 x 2 elements)) elements
    int[][][][] fourDims = new int[][][][] { threeDims, threeDims, threeDims };
    // Two dimension array, conversion only works one way
    Object res = otjc.serialize(fourDims, null, SerializeOptions.DEFAULT);
    OpenType<?> type = ArrayType.getPrimitiveArrayType(int[][][][].class);
    int[][][][] tab = (int[][][][]) converter.deserialize(type, res);
    assertEquals(otjc.serialize(fourDims, null, SerializeOptions.DEFAULT),
        otjc.serialize(tab, null, SerializeOptions.DEFAULT));
}
```

Issue #250, opened 2016-03-17
Option to remove POST request from response


ThomasRasmussen:
Hi

We're about to use Jolokia to provide easy access to JVM statistics, and we are using POST interface (because we need to proxy to JMX) and the JMX listening requires authentication. When using ie. this 

`{"type": "read", "config": { }, "target": {"url": "service:jmx:rmi:///jndi/rmi://localhost:9090/jmxrmi", "password": "secretPassword", "user": "SecretUser"}, "mbean": "java.lang:type=MemoryPool,name=*", "attribute": "Usage" }`

Then the response contains the request including username and password. Since we are putting the output into Splunk, then we really would like to prevent the username/password from being printed.

Is it possible to disable this in the current (1.3.3) version or is it something that can be changed?

Thomas

rhuss:
Unfortunately this is currently not possible since the request is [added always to the response](https://github.com/rhuss/jolokia/blob/master/agent/core/src/main/java/org/jolokia/backend/BackendManager.java#L431-L431), which is use for correlation e.g. in bulk request.

But I agree that this is a valid use case so that we should add this for one of the following versions.

PR #717, opened 2024-07-19
Issue 250

grgrzybek:


tadayosi:
Fix #250

tadayosi (commented):
Hmm, it's not clear to me how the fix addresses the problem stated by Roland:
https://github.com/jolokia/jolokia/issues/250#issuecomment-206513284
> Unfortunately this is currently not possible since the request is [added always to the response](https://github.com/rhuss/jolokia/blob/master/agent/core/src/main/java/org/jolokia/backend/BackendManager.java#L431-L431), which is use for correlation e.g. in bulk request.

Maybe you can add a test for bulk requests with this option to verify it's working with them.

grgrzybek:
I can add a test showing it doesn't simply work ;)
I will mention more explicitly that if you turn this option on, you have to correlate using different method (by order of responses).

tadayosi:
@grgrzybek Yes, I think it's very important as bulk request is a quite common method to optimise Jolokia requests.

grgrzybek:
Oh - I wrote a test which failed... I wanted to ensure that with `includeRequest=false` we're not getting the request inside the response. But it was there!

The point is that `org.jolokia.client.request.ValidatingResponseExtractor#extract()` method is used at client side and it's actually putting the request back into the response - whether or not the response contains its request...

So this confirms my suspicion that we can correlate simply by iterating requests and responses in parallel.

src/documentation/manual/modules/ROOT/pages/client/java.adoc
```diff
@@ -360,6 +360,13 @@ they are created. The allowed values are either
 order (== alphabetical sorted) is used or
 `false` for getting the keys as
 registered. Default is `true`
+
+|`INCLUDE_REQUEST`
+|Whether the response object should contain related request object. +
+This option may be configured globally and overriden at request time.
+When `true`, bulk responses have to be correlated with requests by matching
```
tadayosi, line 367: Maybe it means `false` here?

src/documentation/manual/modules/ROOT/pages/client/javascript.adoc
```diff
@@ -251,6 +251,13 @@ reads.
 |`canonicalNaming`
 |Defaults to true for canonical (sorted) property lists on object names; if set to "false" then they are turned in their unsorted format.
 
+|`includeRequest`
+|Whether the response object should contain related request object. +
+This option may be configured globally and overriden at request time.
+When `true`, bulk responses have to be correlated with requests by matching
```
tadayosi, line 257: `true` -> `false`

src/documentation/manual/modules/ROOT/pages/jolokia_protocol.adoc
```diff
@@ -561,7 +561,10 @@ Only `text/plain` and `application/json` are allowed.
 For any other value Jolokia will fallback to `text/plain`.
 
 `includeRequest`:: A flag (defaults to `true`) which controls whether the incoming
-request should be included in the response (under `request` key).
+request should be included in the response (under `request` key). +
+When `true`, bulk responses have to be correlated with requests by matching
```
tadayosi, line 565: `true` -> `false`

tadayosi (approved):
Some minor typos but otherwise LGTM.

src/documentation/manual/modules/ROOT/pages/jolokia_protocol.adoc
grgrzybek: thanks for detailed review!

PR #719, opened 2024-07-24
Datetime improvements

grgrzybek:


src/documentation/manual/modules/ROOT/pages/jolokia_protocol.adoc
```diff
@@ -656,10 +656,12 @@ converted into a JSON object with the single key
 * `java.net.URL` becomes a JSON
 object with the key `url` containing
 the URL as String.
-* `java.util.Date` is represented in
+* `java.util.Date` is represented by default in
 an ISO-8601 format. When used with a path
 `time` the milliseconds since 1.1.1970
-00:00 UTC are returned.
+00:00 UTC are returned. +
```
tadayosi, line 662: looks like this doesn't hold true anymore as the default is now a local time. shouldn't we instead keep the default time zone to UTC?

tadayosi: sorry, my mistake. please ignore it.

tadayosi (approved):

PR #750, opened 2024-10-07
A Flaky Test Fixed

Derrick2000:
**Issue**:
There is a "flaky test" ([definition](https://www.browserstack.com/test-observability/features/test-reporting/what-is-flaky-test#:~:text=A%20flaky%20test%20refers%20to,challenges%20for%20software%20development%20teams)) found 

`String resp = EnvTestUtil.readToString(url.openStream());`
The string value of `resp` parsed from the stream may have an inconsistent order, which could result in occassional failure in unit tests.

**Observation**:
`assertTrue(resp.matches(".*type.*version.*" + Version.getAgentVersion() + ".*"));`
Our assumption in this test case was that the version number would be written after a type version. However, this order is not guaranteed.

Case 1: (Fail)
image

Case 2: (Fail)
image

Case 3: (Success)
image

**Improvement**
Change the test case into:
`assertTrue(resp.matches(".*type.*version.*") && resp.matches(".*" + Version.getAgentVersion() + ".*"));`
where we ignore the order

### DIFF
--- a/support/spring/src/test/java/org/jolokia/support/spring/BaseServerTest.java
+++ b/support/spring/src/test/java/org/jolokia/support/spring/BaseServerTest.java
@@ -35,7 +35,7 @@ class BaseServerTest {
             URL url = new URL(server.getUrl());
             System.out.println(">>> URL to check: " + server.getUrl());
             String resp = EnvTestUtil.readToString(url.openStream());
-            assertTrue(resp.matches(".*type.*version.*" + Version.getAgentVersion() + ".*"));
+            assertTrue(resp.matches(".*type.*version.*") && resp.matches(".*" + Version.getAgentVersion() + ".*"));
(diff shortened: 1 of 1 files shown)


grgrzybek (approved):

grgrzybek:
Thanks for the PR!

tadayosi (approved):

Derrick2000:
Hi, have I signed off my commit correctly? 
image

grgrzybek:
> Hi, have I signed off my commit correctly? Or Do I need to sign with a GPG key. Thx image

Yes please. See https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification

grgrzybek:
Thanks - now it's fine ;)

PR #782, opened 2025-01-10
780: Kubernetes client: switch from using OkHttp directly to use fabric8 http abstraction

skarsaune:
Issue https://github.com/jolokia/jolokia/issues/780 revealed that credentials were no longer passed on to Jolokia while running over Kubernetes proxy. The solution was to hook into the fabric8 client at a higher abstraction level, so that token refresh interceptor was applied. 

A lot more test coverage was also addedy by using wiremock to simulate both Kubernetes API as well as Jolokia responses over it, ensuring that e.g. credentials are passed correctly.

Note: I noticed that the formatting appers to be inconsistent with some other jolokia modules. I can fix that in a separate PR if desired, but would like to keep the diff minimal as of now.


client/kubernetes/src/test/java/org/jolokia/kubernetes/client/ManualConnectionTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.testng.annotations.Ignore;
+import org.testng.annotations.Test;
+
+/**
+ * This is a utility to attempt to connect to manually test connection to a JVM running in kubernetes
+ *  1. Ensure that your current kubernetes context is set to the context you attempt to use
+ *     (e.g. kubectl config use-context my-context)
+ *  2. Ensure that your kubernetes config contains valid credentials for the k8s cluster of that context
+ */
+public class ManualConnectionTest {
```
skarsaune, line 17: Renamed file to be able to remove the suppresswarnings. It is debatable how useful this manual test is. It might have been a main method as well. However running as a test from the IDE with the ability to use assertions etc. can be nice.

grgrzybek: +1 for keeping it as test instead if `main()`. If the warnings are gone, I'm fine ;)

grgrzybek:
@skarsaune are the files under `src/test/resources/__files` some kind of wiremock convention? I don't see an explicit reference to `__files`...

skarsaune:
> @skarsaune are the files under `src/test/resources/__files` some kind of wiremock convention? I don't see an explicit reference to `__files`...

@grgrzybek  Yes, it is a convention. I could not find the exact reference for test, but it follows the [folder structure described for the server](https://wiremock.org/docs/standalone/java-jar/#json-file-configuration)

grgrzybek:
Thanks - may I just ask you to change tabs to spaces? :)
image

then I can simply merge your commit instead of pushing mine ;)

grgrzybek:
Thanks Martin - I'm releasing Pax URL, Pax Logging and Pax Web today, but tomorrow I'll handle this one.

grgrzybek:
done - many thanks!

PR #784, opened 2025-01-10
Ip6

grgrzybek:


agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServer.java
```diff
@@ -339,25 +339,46 @@ private String detectAgentUrl(HttpServer pServer, JolokiaServerConfig pConfig, S
...
-                    } catch (UnknownHostException e1) {
-                        // Ok, ok. We take the original one
-                        realAddress = serverAddress.getAddress();
-                    }
-                }
-            }
             port = serverAddress.getPort();
         } else {
             realAddress = pConfig.getAddress();
             port = pConfig.getPort();
```
tadayosi, line 345: Just wondering what would happen if the address uses a default port (e.g. 80 or 443) implicitly?  Don't we need to consider a case when the value of port returns `-1`?

server/core/src/main/java/org/jolokia/server/core/util/NetworkInterfaceAndAddress.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public final NetworkInterface networkInterface;
+    public final InetAddress address;
+
+    public NetworkInterfaceAndAddress(NetworkInterface networkInterface, InetAddress address) {
+        this.networkInterface = networkInterface;
+        this.address = address;
+    }
+
+}
```
tadayosi, line 31: Time to consider introducing Java record? :-)  Just an idea...

service/discovery/src/main/java/org/jolokia/service/discovery/DiscoveryMulticastResponder.java
```diff
@@ -20,40 +17,39 @@ public class DiscoveryMulticastResponder extends AbstractJolokiaService<JolokiaS
 
     // Listener threads responsible for creating the response as soon as a discovery request
     // arrives.
-    private final List<MulticastSocketListenerThread> listenerThreads;
+    private MulticastSocketListenerThread listenerThread;
```
tadayosi, line 20: So you finally concluded that a single thread is enough for the class and that's why you removed the usage of List here?  Just checking.

tadayosi (approved):
Overall, great work!

tadayosi:
Btw, isn't it big enough for justifying Jolokia 2.2.0? :-)

rhuss:
Great work, @grgrzybek ! Unfortunately I have no time for an in-depth review :-( , so I leave it to you both :)

server/core/src/main/java/org/jolokia/server/core/util/NetworkInterfaceAndAddress.java
grgrzybek: Hold your horses - we've only recently moved away from Java 8 ;)
But I get your point - would be great!

agent/jvm/src/main/java/org/jolokia/jvmagent/JolokiaServer.java
grgrzybek: For `InetAddress` created explicitly indeed we'd have to handle it. But here, this particular `InetAddress` is obtained from `com.sun.net.httpserver.HttpServer#getAddress()`, so we're fine.
Thanks for highlighting!

service/discovery/src/main/java/org/jolokia/service/discovery/DiscoveryMulticastResponder.java
grgrzybek: Exactly. I realized (with my friend debugger) that in my case where I have `eth0` and `tun0` interfaces, each with 2x IPv6 addresses and 1x IPv4 address, I had **six** `MulticastSocketListenerThread` threads, each listening on _reusable_ multicast UDP socket - each on **the same** `*:24884` interface:port.

Issue #796, opened 2025-02-04
Jolokia 2.2.1 is extremely slow at start on macOS due to newly introduced reverse DNS lookups


tadayosi:
You can reproduce the issue by running the following `hello.java` with jbang:

**hello.java**
```java
public class hello {
    public static void main(String... args) {
        System.out.println("Hello World");
    }
}
```

```console
$ time jbang --javaagent=org.jolokia:jolokia-agent-jvm:2.2.1:javaagent hello.java
Hello World
jbang --javaagent=org.jolokia:jolokia-agent-jvm:2.2.1:javaagent hello.java
  0.18s user 0.09s system 0% cpu 45.548 total
```

It took 45s on my MacBook Pro M3.

Thread dump shows that it's waiting at `java.net.InetAddress.getHostName()` which causes a reverse DNS lookup for unresolvable IPv6 address.
```
"main" #1 [3587] prio=5 os_prio=31 cpu=59.17ms elapsed=20.49s tid=0x000000012b00a400 nid=3587 runnable  [0x000000016b609000]
   java.lang.Thread.State: RUNNABLE
	at java.net.Inet6AddressImpl.getHostByAddr(java.base@21.0.6/Native Method)
	at java.net.InetAddress$PlatformResolver.lookupByAddress(java.base@21.0.6/InetAddress.java:1225)
	at java.net.InetAddress.getHostFromNameService(java.base@21.0.6/InetAddress.java:840)
	at java.net.InetAddress.getHostName(java.base@21.0.6/InetAddress.java:782)
	at java.net.InetAddress.getHostName(java.base@21.0.6/InetAddress.java:754)
	at org.jolokia.server.core.config.StaticConfiguration.lambda$initializeFromNetwork$1(StaticConfiguration.java:177)
	at org.jolokia.server.core.config.StaticConfiguration$$Lambda/0x0000000801006210.accept(Unknown Source)
	at java.util.HashMap.forEach(java.base@21.0.6/HashMap.java:1429)
	at org.jolokia.server.core.config.StaticConfiguration.initializeFromNetwork(StaticConfiguration.java:166)
	at org.jolokia.server.core.config.StaticConfiguration.<init>(StaticConfiguration.java:142)
	at org.jolokia.jvmagent.JolokiaServerConfig.init(JolokiaServerConfig.java:121)
	at org.jolokia.jvmagent.JvmAgentConfig.<init>(JvmAgentConfig.java:62)
[...]

PR #798, opened 2025-02-04
[#796] Categorize network interfaces also by the number of assigned a…

grgrzybek:
…ddresses

### DIFF
--- a/server/core/src/main/java/org/jolokia/server/core/util/NetworkUtil.java
+++ b/server/core/src/main/java/org/jolokia/server/core/util/NetworkUtil.java
@@ -16,8 +16,10 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * Utility class for network related stuff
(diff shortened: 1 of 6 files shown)


grgrzybek:
What do you think about this @tadayosi ?

tadayosi (approved):

server/core/src/main/java/org/jolokia/server/core/util/NetworkUtil.java
```diff
@@ -28,7 +30,7 @@
 public final class NetworkUtil {
 
     private static final boolean MAC = System.getProperty("os.name").toLowerCase().contains("mac");
-    private static final List<String> MAC_SKIP_NIFS = List.of("awdl0", "llw0");
+    private static final List<String> MAC_SKIP_NIFS = List.of("awdl", "llw");
```
tadayosi, line 33: If we go this approach, we can include `utun` here as well. Then a single place where `name.startsWith("utun")` is used should be no longer necessary.

grgrzybek: What if someone wants to bind Jolokia on `utun`? :)

grgrzybek:
PR updated. @tadayosi may I ask you to use:
```java
private static final List<String> MAC_SKIP_NIFS = Collections.emptyList();
```

locally and check how it works now? (Hoping that usage of `allowDnsReverseLookup` is enough).

server/core/src/main/java/org/jolokia/server/core/util/NetworkUtil.java
tadayosi: Who wants? :-)

tadayosi:
@grgrzybek Yes, it works too. But it's still useful when testing Jolokia on Mac. It boosts the speed of the test for Mac. I'll change it to empty but set it up only from tests, so that in production it doesn't exclude `utun`.

tadayosi (approved):

PR #801, opened 2025-02-07
789: Deeper testing, and found a few fields that should be integer

skarsaune:
Changing to default to `java.lang.Long` largely fixed binary compatibility issues between local/RMI connections and jolokia. 
I added some tests do do deeper type testing (We should be aware about complexity , brittleness (if any) and maintainability of these tests ...) .

It actually uncovered 3 cases that should be integer instead of long.

cc: @grgrzybek


grgrzybek:
Hmm, d3ce97e288c4e532355a98a4e448540c55db1598 was about losing numeric type when doing JSON conversion.. I'm not sure what's the impact for client/jmx-adapter...

grgrzybek:
Give me some time to review - today is not a great day...

tadayosi (approved):

PR #809, opened 2025-03-04
Utility functions to help analysing jolokia responses

phantomjinx:
* Adds tests for utility functions

Tracked issue: https://github.com/hawtio/hawtio-online/issues/396

### DIFF
--- a/client/javascript-esm/packages/jolokia-simple/src/jolokia-simple-types.ts
+++ b/client/javascript-esm/packages/jolokia-simple/src/jolokia-simple-types.ts
@@ -41,6 +41,14 @@ interface JolokiaSimpleStatic extends JolokiaStatic {
 
   /** Creating {@link IJolokiaSimple} using object creation syntax with `new` */
   new(config: JolokiaConfiguration | string): IJolokiaSimple
+
+  /**
+   * Utility method which checks whether a response is a version response
+   * @param resp response to check
+   * @return true if response is a success
+   */
+  isVersionResponse(resp: unknown): resp is VersionResponseValue
+
 }
 
 /**
@@ -155,6 +163,13 @@ interface IJolokiaSimple extends IJolokia {
    * @param params optional path and parameters to be passed to `Jolokia.request()` (params is the last object argument)
    */
   list(...params: (string[] | string | SimpleRequestOptions)[]): Promise<ListResponseValue>
+
+  /**
+   * Utility method which checks whether a response is a version response
+   * @param resp response to check
+   * @return true if response is a success
+   */
+  isVersionResponse(resp: unknown): resp is VersionResponseValue
 }
 
 /**
--- a/client/javascript-esm/packages/jolokia-simple/src/jolokia-simple.ts
+++ b/client/javascript-esm/packages/jolokia-simple/src/jolokia-simple.ts
@@ -276,6 +276,11 @@ Jolokia.prototype.list = async function(this: IJolokia, ...params: (string[] | s
     })
 }
 
+Jolokia.prototype.isVersionResponse = function (resp: unknown): resp is VersionResponseValue {
+  if (!resp || typeof resp !== 'object') return false
+  return 'protocol' in resp && 'agent' in resp
+}
+
 // ++++++++++++++++++++++++++++++++++++++++++++++++++
 // Private/internal functions
 
--- a/client/javascript-esm/packages/jolokia-simple/test/jolokia-simple.test.ts
+++ b/client/javascript-esm/packages/jolokia-simple/test/jolokia-simple.test.ts
@@ -130,4 +130,15 @@ describe("Jolokia simple API", () => {
     expect(Object.keys((v as ListResponseValue)!)[0]).toBe("JMImplementation")
   })
 
+  test("Simple Version Response", async () => {
+    const jolokia = new Jolokia({ url: `http://localhost:${port}/jolokia-simple` }) as IJolokiaSimple
+    const v = await jolokia.version()
+    expect(jolokia.isVersionResponse(v)).toBe(true)
+  })
+
+  test("Simple Not a Version Response", async () => {
+    const jolokia = new Jolokia({ url: `http://localhost:${port}/jolokia-simple` }) as IJolokiaSimple
+    const v = await jolokia.list()
+    expect(jolokia.isVersionResponse(v)).toBe(false)
+  })
 })
--- a/client/javascript-esm/packages/jolokia/src/jolokia-types.ts
+++ b/client/javascript-esm/packages/jolokia/src/jolokia-types.ts
@@ -927,6 +927,20 @@ interface JolokiaStatic {
    * @return true if response is an error, false otherwise
    */
   isError(resp: JolokiaResponse): resp is JolokiaErrorResponse
+
+  /**
+   * Utility method which checks whether a response is success response (HTTP perspective)
+   * @param resp response to check
+   * @return true if response is a success
+   */
+  isResponseSuccess(resp: unknown): resp is JolokiaSuccessResponse
+
+  /**
+   * Utility method which checks whether a response is an error response (HTTP perspective)
+   * @param resp response to check
+   * @return true if response is a success
+   */
+  isResponseError(resp: unknown): resp is JolokiaErrorResponse
 }
 
 /**
--- a/client/javascript-esm/packages/jolokia/src/jolokia.ts
+++ b/client/javascript-esm/packages/jolokia/src/jolokia.ts
@@ -487,6 +487,16 @@ Jolokia.isError = Jolokia.prototype.isError = function (resp: JolokiaResponse): 
   return resp == null || resp.status != 200
 }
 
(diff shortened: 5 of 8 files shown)


phantomjinx:
Wonder which way to go here ...

The Jolokia library exports `Jolokia` as a default. By including the utility functions in the library, a mix of named and default exports causes issues - rollup complains:
```
Consumers of your bundle will have to use chunk.default to access their default export, which may not be what you want. Use `output.exports: "named"` to disable this warning.
```
More importantly, jolokia-simple fails to find `Jolokia.prototype` and so errors occur in test apps.

I see no reason to change the default as the utility functions are not important enough to move to named exports. So wondering how else to go.

For example, should I package the exports as a sub name listed in `packages.json`, eg.
```
  "exports": {
    ".": {
      "types": "./dist/dts/jolokia.d.ts",
      "import": "./dist/jolokia.mjs",
      "default": "./dist/jolokia.cjs"
    },
    "./utils": {
      "types": "./dist/dts/jolokia-utils.d.ts",
      "import": "./dist/jolokia-utils.mjs",
      "default": "./dist/jolokia-utils.cjs"
    }
  }
```

@grgrzybek WDYT??

grgrzybek:
```json
    "./utils": {
      "types": "./dist/dts/jolokia-utils.d.ts",
      "import": "./dist/jolokia-utils.mjs",
      "default": "./dist/jolokia-utils.cjs"
    }
```

looks good to me, but does it still cause Rollup errors?

tadayosi (commented):
I think instead of migrating all of them to jolokia.js we can migrate each function one by one. It's because some functions don't fit to jolokia.js, for example, `isObject` is too generic and maybe is not justified to have it in jolokia.js. OTOH, it's not clear why we need `responseParse` in jolokia.js. It seems to be only expected to be used for hawtio-online (even hawtio-react doesn't seem to need it).
I think the following functions are valid to have in jolokia.js:
* isResponseSuccess
* isResponseError
* isVersionResponse

However, note that the main Jolokia object already has a similar util function, so instead of having these utils in a separate file, I rather expect them to be added to the Jolokia object itself. And for `isVersionResponse` it should be added to the jolokia-simple.ts.
https://github.com/jolokia/jolokia/blob/a19090bd0eda2356a98ca92343a0ff4371c4c8c2/client/javascript-esm/packages/jolokia/src/jolokia.ts#L486-L488

client/javascript-esm/packages/jolokia/src/jolokia-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+} from "./jolokia-types.js"
+
+export type ParseResult<T> = { hasError: false; parsed: T } | { hasError: true; error: string }
+
+function isObject(value: unknown): value is object {
+  const type = typeof value
+  return value != null && (type === 'object' || type === 'function')
+}
+
+export function isResponseSuccessType(o: unknown): o is JolokiaSuccessResponse {
```
tadayosi, line 14: The function name seems to be a bit redundant. Considering we already have `isError(resp: JolokiaResponse)`, maybe `isResponseSuccess(o: unknown)` is better.

client/javascript-esm/packages/jolokia/src/jolokia-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+  return isObject(o) && 'status' in o && 'timestamp' in o && 'value' in o
+}
+
+export function isResponseErrorType(o: unknown): o is JolokiaErrorResponse {
```
tadayosi, line 18: So we already have `isError(resp: JolokiaResponse)`. The only difference and its value proposition should be the type of the argument, where the original is `JolokiaReponse` while the new function is `unknown`.

We can also shorten the name as it's a bit redundant. Maybe `isResponseError(o: unknown)`?

client/javascript-esm/packages/jolokia/src/jolokia-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+  return isObject(o) && 'error_type' in o && 'error' in o
+}
+
+export function isVersionResponseType(o: unknown): o is VersionResponseValue {
```
tadayosi, line 22: I think we can extend the main Jolokia object from `jolokia-simple.ts` for the function. And again the name seems a bit redundant, maybe we can name it `isVersionResponse(o: unknown)`.

phantomjinx:
Adding them to the prototype would solve a number of problems.

Turns out using package.json [exports] turns the build into an ECMAScript module. Importing this into hawtio-online causes a compile error in the form:
```
The current file is a CommonJS module whose imports will produce require() calls; however, the referenced file is an ECMAScript module and cannot be imported via require(). Consider writing a dynamic import() call instead.
```

client/javascript-esm/packages/jolokia/src/jolokia-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * If the body is an error then the ParseResult
+ * - hasError flag is set to true
+ * - error property is populated
+ *
+ * If the body is successfully parsed then the ParseResult
+ * - hasError flag is set to false
+ * - parsed property is populated
+ */
+export async function responseParse(response: unknown): Promise<ParseResult<JolokiaSuccessResponse | JolokiaErrorResponse>> {
```
tadayosi, line 37: > OTOH, it's not clear why we need `responseParse` in jolokia.js.

Now I read the function again. I understand the idea, but firstly we no longer seem to need `async` here, as you removed `response.json()` from the original function. Next, `try`-`catch` is also no longer needed as the code now shouldnt throw any exception.

And after being aware that `response.json()` is removed, the function now just checks whether the given unknown object is success or error. At this point, I don't see there's any generic value in the function. For most Jolokia client apps, we can just repeat the same if-else within the app code to achieve the same thing.

tadayosi (changes requested):

client/javascript-esm/packages/jolokia/src/jolokia-utils.ts
phantomjinx: Yeah. Agree with that.

tadayosi (approved):
Looks great to me!

tadayosi:
@phantomjinx One last request from me is that Jolokia project sets a requirement to accept contributions, which is to sign your commits with GPG. This PR is currently blocked because it's not GPG signed. Please follow the link and sign it. (It's not only about adding your name to the commit message with `git commit -s`).
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

grgrzybek:
Now it looks good ;) We don't have new ts file (utils), however...

> Turns out using package.json [exports] turns the build into an ECMAScript module. Importing this into hawtio-online causes a compile error in the form:
>
> ```
> The current file is a CommonJS module whose imports will produce require() calls; however, the referenced file is an ECMAScript module and cannot be imported via require(). Consider writing a dynamic import() call instead.
> ```

I'm pretty sure that I took care that `jolokia.js` NPM package can be used both as CommonJS and ESM... I remember it was painful to tweak rollup, webpack, jest, package.json and other files to make sure everything works correctly... Maybe there's something in hawtio-online? Remember that Jolokia and Jolokia Simple `package.json` has:
```json
"type": "module"
```

in the first place...

And now about the code.

I see you've added `isResponseSuccess` and `isResponseError` to Jolokia prototype (non-static, instance methods) - similar to existing `isError`, `escape` and `escapePost`. You've also added these as _static methods_ (to `Jolokia` object itself). While this is consistent with what we already have, I'd like to avoid adding these to `Jolokia.prototype` - `isError` is there only for historical reasons - these methods do not rely on state of `Jolokia` _instance_, so having them static should be enough.

So I'd change:
```javascript
Jolokia.isResponseSuccess = Jolokia.prototype.isResponseSuccess = function(...
Jolokia.isResponseError = Jolokia.prototype.isResponseError = function(...
```

to:
```javascript
Jolokia.isResponseSuccess = function(...
Jolokia.isResponseError = function(...
```

and remove these from `IJolokia`. In tests you use:
```javascript
expect(Jolokia.isResponseSuccess(response.body)).toBe(true)
```

and not:
```javascript
const j = new Jolokia(...)
expect(j.isResponseSuccess(response.body)).toBe(true)
```

grgrzybek:
> @grgrzybek
> 
> Changing the `Jolokia.isResponseSuccess`, `Jolokia.isResponseError` to remove the prototypes is fine but what would you want to do about `Jolokia.prototype.isVersionResponse` which is in `jolokia-simple.ts` ? Seems that consistency would lead to this also being on `Jolokia` itself rather than the prototype but that cannot happen without being included in `JolokiaStatic` so it cannot be in `jolokia-simple.ts` but in `jolokia.ts`.

there's `interface JolokiaSimpleStatic extends JolokiaStatic` in `jolokia-simple-types.ts`, so I'd add it there. And while `jolokia-types.ts` define various types related to Jolokia operations (`VersionResponseValue`, `SearchResponseValue`, ...), there are no methods for particular Jolokia operations - only generic ones.

PR #821, opened 2025-04-05
820: Allow information whether http or https was used

skarsaune:


tadayosi (commented):
Looks good but I can only see you refactored code but added no substantial changes.

What's this pull req really supposed to do?

tadayosi:
Ah, so this pull req is about just exposing the internal logic as the public `getJolokiaProtocol` method?  If so, it looks good to me 👍

grgrzybek:
I delayed the review - sorry for that!
Thanks @tadayosi !

@skarsaune btw what's this `jmx.remote.x.check.stub` property?

client/jmx-adapter/src/main/java/org/jolokia/client/jmxadapter/JolokiaJmxConnector.java
```diff
@@ -71,6 +67,15 @@ public void connect(Map<String, ?> env) throws IOException {
     postCreateAdapter();
   }
 
+  public String getJolokiaProtocol(Map<String, Object> env) {
```
tadayosi, line 70: In the intention to make this logic public, is it clear to the caller what the argument `env` should really be?  It should be the merged env but I'm not sure if it's really a shared assumption with the users. Rather, isn't it better to make it a boolean arg `checkStub` or something like that so that the users won't need the internal details?

Also, adding javadoc for the method should definitely help.

tadayosi:
@grgrzybek

> @skarsaune btw what's this `jmx.remote.x.check.stub` property?

I can only find this:
https://jolokia.org/reference/html/manual/clients.html#remote-jmx-example

skarsaune:
Sorry for the late reply guys. Back in the day it was based on testing with jvisualvm that had a setting: 

https://github.com/oracle/visualvm/issues/267

Perhaps, as you say, it would have been better with an explicit setting to force HTTPS, however currently we do not have such a concept in mission control (as I understand it). that is another problem though, and not jolokias problem. (Will dig up the reference later)

grgrzybek (approved):

grgrzybek:
@skarsaune thanks for the PR - now I understand - later than @tadayosi (and not only because of the timezone difference ;).

I still have plans to review integration with visualvm, but didn't have time to do it thoroughly...

The last missing piece here (for the sake of the process) is signed commit - Martin, may I ask you for `git commit -S` here? See https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #841, opened 2025-06-11
Fix SerializerTest on Windows

apupier:


grgrzybek:
A tiny suggestion @apupier - can we do this without `commons-text`? :)

grgrzybek (approved):
Thanks!

grgrzybek:
It's blocked - in Jolokia there's a commit policy that requires `git commit -S`.
See https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key

PR #852, opened 2025-07-10
Remove deprecated Pragma header

GChuf:
Pragma header is deprecated - Cache-Control is enough.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Pragma

Let me know if any other changes are needed.


grgrzybek (approved):
:+1: Thanks @GChuf !

GChuf:
> May I ask you for `git commit -S`? See https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Done, finally!

PR #909, opened 2025-10-02
Added more attributes to ATTRIBUTES_NOT_SAFE_FOR_DIRECT_COMPARISON

lawrencewang49:
## Summary
This PR updates the testMBeanInfo test in JmxBridgeTest by adding three more attributes to `ATTRIBUTES_NOT_SAFE_FOR_DIRECT_COMPARISON`.

Previously, the test compared all MBean attribute values between the native MBean server and the Jolokia adapter. Some attributes represent live or dynamic JVM data (e.g., current thread counts, loaded class counts, logger names) that can change between invocations. Direct 1:1 comparison on these attributes led to nondeterministic test failures.

---

## Changes made
Added the following attributes to the list of attributes not safe for direct comparison:
```java
"java.util.logging:type=Logging.LoggerNames",
"java.lang:type=ClassLoading.LoadedClassCount",
"java.lang:type=Threading.PeakThreadCount"
```
These attributes are now skipped during direct value comparison in testMBeanInfo.

---

## Rationale
- Prevents false negatives: The values of these attributes vary dynamically during test execution, so comparing them directly could cause nondeterministic failures.
- Maintains test focus: The test still validates the structural correctness and static attributes of MBeans, without being broken by live runtime data.
- Improves stability: Ignoring these specific live attributes ensures that the test suite passes consistently across different JVM versions and runtime conditions.


grgrzybek:
Thanks!

Issue #951, opened 2025-12-04
IPv6 addresses in JolokiaJmxConnector give IllegalArgumentException


tommystendahl:
Hi,
I think I have found a problem using IPv6 addresses with JolokiaJmxConnector, this results in a IllegalArgumentException. I have created a test case to show the issue: https://github.com/tommystendahl/jolokia/commit/1b9b14955aaa2b995379cf9b766021dfe2ad4dd9

It looks like the [] around the IPv6 addresses are strip away here: https://github.com/jolokia/jolokia/blob/b3aea9fc97bcd51d6303537fa6db65d7fbb74535/client/jmx-adapter/src/main/java/org/jolokia/client/jmxadapter/JolokiaJmxConnector.java#L59

The JMXServiceURL.getHost() returns the host without [] and this will result in a IllegalArgumentException.

I also did a patch to fix the issue: https://github.com/tommystendahl/jolokia/commit/af20cac694287370125d7306ea722579e998747d

PR #952, opened 2025-12-04
IPv6 addresses in JolokiaJmxConnector give IllegalArgumentException

tommystendahl:
issue #951


grgrzybek:
Thanks for the PR - looks good!
I plan to release 2.5.0 this year, so this should be included.
Please (if possible) change this PR to contain `git commit -S` kind of commit (with signature). If you don't have/want GPG keys to be used, I'll add this commit anyway.

tommystendahl:
Thanks for the quick response. 2.5.0 this year would be great.
I will see what I can do about the signatures on the commits.

tommystendahl:
I updated with a new commit that is signed. I put it in one commit but there are no changes to the code.

Issue #978, opened 2026-02-23
Can't find org.springframework.boot.actuate.autoconfigure.endpoint.expose.EndpointExposure


cunningt:
Getting the following stack from one of the camel-jolokia-starter tests (https://github.com/apache/camel-spring-boot/tree/main/components-starter/camel-jolokia-starter)

I think spring-boot-autoconfigure in jolokia-support-springboot should be compile scope so that we can find org.springframework.boot.actuate.autoconfigure.endpoint.expose.EndpointExposure

```
  [ERROR] org.apache.camel.itest.springboot.CamelJolokiaTest.componentTests -- Time
  elapsed: 0.752 s <<< ERROR!
  IdentifiedTestException [collectedExceptions={[engine:junit-jupiter]/[class:org.ap
  ache.camel.itest.springboot.CamelJolokiaTest]/[method:componentTests()]=java.lang.
  IllegalArgumentException: Could not find class [org.springframework.boot.actuate.a
  utoconfigure.endpoint.expose.EndpointExposure]}]
      at org.jboss.arquillian.junit5.container.JUnitJupiterTestRunner$ArquillianTest
  MethodExecutionListener.getTestResult(JUnitJupiterTestRunner.java:113)
      at org.jboss.arquillian.junit5.container.JUnitJupiterTestRunner$ArquillianTest
  MethodExecutionListener.access$100(JUnitJupiterTestRunner.java:72)
      at org.jboss.arquillian.junit5.container.JUnitJupiterTestRunner.execute(JUnitJ
  upiterTestRunner.java:61)
      at org.jboss.arquillian.protocol.jmx.JMXTestRunner.doRunTestMethod(JMXTestRunn
  er.java:152)
      ...
      at org.springframework.context.annotation.ConfigurationClassParser.parse(Confi
  gurationClassParser.java:200)
      at org.springframework.context.annotation.ConfigurationClassPostProcessor.proc
  essConfigBeanDefinitions(ConfigurationClassPostProcessor.java:446)
      at org.springframework.context.annotation.ConfigurationClassPostProcessor.post
  ProcessBeanDefinitionRegistry(ConfigurationClassPostProcessor.java:310)
      at org.springframework.context.support.PostProcessorRegistrationDelegate.invok
  eBeanDefinitionRegistryPostProcessors(PostProcessorRegistrationDelegate.java:349)
      at org.springframework.context.support.PostProcessorRegistrationDelegate.invok
[...]

PR #979, opened 2026-02-23
Remove provided scope from spring-boot-autoconfigure; upgrade spring to 7.0.5 / spring-boot to 4.0.3

cunningt:
https://github.com/jolokia/jolokia/issues/978

Remove the scope from the spring-boot-autoconfigure dependency; needed to find org.springframework.boot.actuate.autoconfigure.endpoint.expose.EndpointExposure

Also, upgrade spring / spring-boot to latest compatible releases


grgrzybek:
Hmm, is this the fix for #978? I remember adding `provided` scope on purpose...
Let me check.

grgrzybek:
Closed as commented under #978.

PR #1010, opened 2026-05-29
JOLOKIA-989: Eliminate need for apache http 4 in kubernetes adapter

skarsaune:


grgrzybek (approved):
Looks good!
Thanks @skarsaune

grgrzybek:
May I ask for a `git commit -S`?

grgrzybek:
Thanks @skarsaune again.
I'll think about new release soon.
I'm currently investigating how Jolokia could use token auth at JVM agent level without too many external dependencies. If the investigation takes too long, I'll release 2.6.1 with this change only.

skarsaune:
@grgrzybek  : On second thought we should probably rename the MinimalHttpAdapter to something that aligns better with the other http spi implementations. I can think about a separate follow up PR.

grgrzybek:
Sure - I won't volunteer to pick up the name ;)

Issue #143, opened 2014-06-14
Enhanced CORS support


rhuss:
For proper CORS support with authentication the OPTIONS preflight request must happen unauthenticated. See https://gist.github.com/rhuss/e0c949f7564a47943460 for details and for how to fix this.

rhuss:
This is delayed for Jolokia 2.0. Too big for 1.x.

grgrzybek:
Almost done. I'll soon send a PR with the changes and updated documentation, but to summarize:
* `OPTIONS` request that is identified as actual preflight request (i.e., has `Origin` and `Access-Control-Request-Method` headers) is never autthenticated - both in WAR and JVM deployments (I'll check Spring deployments too)
* Proper response headers are sent - both for preflight and actual requests - taking into account the allowed origins from the policy restrictor. Also `Access-Control-Allow-Credentials` is sent according to the configured authentication (if no authentication is configured, the header is sent with `false` value)

Here's an example of a simulated preflight request:
```
$ curl -XOPTIONS -H'Access-Control-Request-Method: POST' -H'Origin: http://localhost:3000' -i http://localhost:7778/jolokia
HTTP/1.1 200 OK
Access-control-max-age: 7200
Date: Thu, 02 Jul 2026 10:13:04 GMT
Access-control-allow-methods: POST, GET
Access-control-allow-origin: http://localhost:3000
Access-control-allow-credentials: true
Content-length: 0
```

and the same for a request that is _not_ a preflight request when authentication is enabled:
```
$ curl -XOPTIONS -H'Origin: http://localhost:3000' -i http://localhost:7778/jolokia
HTTP/1.1 401 Unauthorized
Www-authenticate: Basic realm="jolokia", charset="UTF-8"
Date: Thu, 02 Jul 2026 10:11:41 GMT
Access-control-allow-origin: http://localhost:3000
Access-control-allow-credentials: true
Content-length: 0
```

and when authentication is disabled:
```
$ curl -XOPTIONS -H'Origin: http://localhost:3000' -i http://localhost:7778/jolokia
HTTP/1.1 400 Bad Request
Date: Thu, 02 Jul 2026 10:12:34 GMT
Content-type: text/plain; charset=utf-8
Content-length: 86

400 (Bad Request)

HTTP Method OPTIONS is supported only for CORS preflight requests.
```

grgrzybek:
Additionally we send `Vary: Origin` response header whenever the response depends on the incoming `Origin` header

PR #1021, opened 2026-06-30
Json escaping and serializing improvements

GChuf:
This comes after a series of optimizations on apache artemis for ArtemisMBeanServerGuard.canInvoke (basically for jolokia's list request)
After improvements on that side, jolokia's json escaping becomes one of the lowest hanging fruits for speeding up (it accounts for about 5-10% of the cpu time).

---
There's also a bug fix according to gemini:

The firstimplementation contains a mathematical bug when escaping control characters between 0x10 and 0x1F (e.g., Data Link Escape, Device Controls). Look at how it maps the hex values to characters:
```
buffer.append((char)(((c & 0xf0) >> 4) + '0')); // For 0x10-0x1F, this evaluates to '1'
buffer.append((char)((c & 0x0f) + '0'));        // Bug here
```

If c = 0x1A (ASCII 26), the second line evaluates to 0x0A + '0', which is 10 + 48 = 58. In ASCII, 58 is the colon character (:), not a hex letter.
```
Implementation 2 output: \u001a (Correct)
Implementation 1 output: \u001: (Malformed JSON)
```
---

### DIFF
--- a/json/src/main/java/org/jolokia/json/JSONWriter.java
+++ b/json/src/main/java/org/jolokia/json/JSONWriter.java
@@ -36,7 +36,7 @@ public class JSONWriter {
         writer.write('{');
         int pos = map.size() - 1;
         for (Map.Entry<String, Object> el : map.entrySet()) {
-            escape(writer, el.getKey().toCharArray());
+            escape(writer, el.getKey());
             writer.write(':');
             serialize(el.getValue(), writer);
             if (pos-- > 0) {
@@ -93,31 +93,32 @@ public class JSONWriter {
     public static void serialize(Object value, Writer writer) throws IOException {
         if (value == null) {
             writer.write("null");
-        } else if (value instanceof Boolean) {
-            writer.write((boolean) value ? "true" : "false");
-        } else if (value instanceof Float) {
-            if (Float.isFinite((float) value)) {
-                writer.write(value.toString());
+        } else if (value instanceof Boolean b) {
+            writer.write(b ? "true" : "false");
+        } else if (value instanceof Number num) {
+            if (num instanceof Double d) {
+                if (Double.isFinite(d)) {
+                    writer.write(d.toString());
+                } else {
+                    writer.write("null");
+                }
+            } else if (num instanceof Float f) {
+                if (Float.isFinite(f)) {
+                    writer.write(f.toString());
+                } else {
+                    writer.write("null");
+                }
             } else {
-                writer.write("null");
+                writer.write(num.toString());
             }
-        } else if (value instanceof Double) {
-            if (Double.isFinite((double) value)) {
-                writer.write(value.toString());
-            } else {
-                writer.write("null");
-            }
-        } else if (value instanceof Number) {
-            // includes BigDecimals and BigIntegers
-            writer.write(value.toString());
-        } else if (value instanceof Character) {
-            escape(writer, new char[] { (char) value });
-        } else if (value instanceof String) {
-            escape(writer, ((String) value).toCharArray());
+        } else if (value instanceof Character c) {
+            escape(writer, c);
+        } else if (value instanceof String s) {
+            escape(writer, s);
         } else if (value instanceof Collection<?> collection) {
             serialize(collection, writer);
-        } else if (value instanceof JSONObject) {
-            serialize((JSONObject) value, writer);
+        } else if (value instanceof JSONObject j) {
+            serialize(j, writer);
         } else if (value instanceof Map<?, ?> map) {
             // not sure about the key types, so be extra careful
             serializeAnyMap(map, writer);
@@ -135,14 +136,15 @@ public class JSONWriter {
     }
 
     /**
-     * When writing string values we have to escape characters. This method writes escaped char array into
-     * the target {@link Writer} but optimizing the process by not writing it one char at a time. Also
-     * the char array is surrounded by quotes.
+     * When writing string values we have to escape characters. This method uses a sliding-window
+     * approach to track contiguous blocks of unescaped characters.
+     * The output is enclosed in double quotes.
      * @param characters
      * @param writer
      * @throws IOException
      */
-    private static void escape(Writer writer, char[] characters) throws IOException {
+
+    private static void escape(Writer writer, String str) throws IOException {
         // https://datatracker.ietf.org/doc/html/rfc8259#section-7
         //     All Unicode characters may be placed within the
         //     quotation marks, except for the characters that MUST be escaped:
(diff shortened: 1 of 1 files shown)


GChuf:
I've also benchmarked the old and the new methods (I've used artemis project since it proved to be the faster route for me to build/execute this jmh test - as artemis already uses jmh)

Results:
### 📊 JMH Benchmark Results

| Benchmark Method | Profile | Mode | Score (ns/op) | Error (± ns/op) |
| :--- | :--- | :---: | :---: | :---: |
| **`legacyEscape0_fromString`** | PLAIN_SHORT | avgt | 85.587 | 8.298 |
| `optimizedEscape_fromString` | PLAIN_SHORT | avgt | **16.967** | 2.372 |
| **`legacyEscape0_singleChars`** | PLAIN_SHORT | avgt | 305.696 | 43.056 |
| `optimizedEscape_singleChars` | PLAIN_SHORT | avgt | **14.305** | 3.042 |
| | | | | |
| **`legacyEscape0_fromString`** | PLAIN_LONG | avgt | 1950.216 | 321.333 |
| `optimizedEscape_fromString` | PLAIN_LONG | avgt | **457.203** | 174.821 |
| **`legacyEscape0_singleChars`** | PLAIN_LONG | avgt | 8603.659 | 2002.909 |
| `optimizedEscape_singleChars` | PLAIN_LONG | avgt | **343.226** | 40.599 |
| | | | | |
| **`legacyEscape0_fromString`** | SPARSE_ESCAPES | avgt | 1336.699 | 293.270 |
| `optimizedEscape_fromString` | SPARSE_ESCAPES | avgt | **508.286** | 97.485 |
| **`legacyEscape0_singleChars`** | SPARSE_ESCAPES | avgt | 12987.695 | 2899.638 |
| `optimizedEscape_singleChars` | SPARSE_ESCAPES | avgt | **593.921** | 148.733 |
| | | | | |
| **`legacyEscape0_fromString`** | DENSE_ESCAPES | avgt | 635.321 | 159.436 |
| `optimizedEscape_fromString` | DENSE_ESCAPES | avgt | **400.519** | 43.576 |
| **`legacyEscape0_singleChars`** | DENSE_ESCAPES | avgt | 6645.354 | 2008.071 |
| `optimizedEscape_singleChars` | DENSE_ESCAPES | avgt | **279.504** | 190.270 |
| | | | | |
| **`legacyEscape0_fromString`** | ALL_ESCAPES | avgt | 696.125 | 149.292 |
| `optimizedEscape_fromString` | ALL_ESCAPES | avgt | **593.518** | 31.902 |
| **`legacyEscape0_singleChars`** | ALL_ESCAPES | avgt | **81.046** | 7.739 |
| `optimizedEscape_singleChars` | ALL_ESCAPES | avgt | 125.531 | 25.147 |


Apart from the speedups - the new implementation uses no allocations, which is also a plus.


The actual test:
```

package org.apache.activemq.artemis.tests.performance.jmh;


import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.Param;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;
import java.io.IOException;
import java.io.Writer;
import java.util.concurrent.TimeUnit;
import org.openjdk.jmh.annotations.OutputTimeUnit;

import org.openjdk.jmh.annotations.Level;


@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Thread)
@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
@Fork(value = 1, jvmArgsAppend = {"-Xms1g", "-Xmx1g"})
public class JsonEscapePerfTest {

    /**
     * Escape-density profiles exercise the two algorithms differently:
     * <ul>
     *   <li>PLAIN_SHORT / PLAIN_LONG - nothing needs escaping, so the new
     *       run-based method should collapse to a single writer.write() call.</li>
     *   <li>SPARSE_ESCAPES - occasional escapes, roughly the common case for
     *       real JSON string values (ids, names, free text).</li>
     *   <li>DENSE_ESCAPES - about 1 in 3 characters needs escaping.</li>
     *   <li>ALL_ESCAPES - every character needs escaping. This is the
     *       adversarial case for the run-based method, since every char
     *       becomes its own writer.write() call instead of one bulk write.</li>
     * </ul>
     */
    @Param({"PLAIN_SHORT", "PLAIN_LONG", "SPARSE_ESCAPES", "DENSE_ESCAPES", "ALL_ESCAPES"})
    private String profile;

    private String testString;
    private char[] testChars;
    private CountingWriter writer;

    @Setup(Level.Trial)
    public void setupTrial() {
        testString = buildTestString(profile);
        testChars = testString.toCharArray();
        writer = new CountingWriter();
    }

    private static String buildTestString(String profile) {
        StringBuilder sb = new StringBuilder();
        switch (profile) {
            case "PLAIN_SHORT" -> sb.append("the quick brown fox");
            case "PLAIN_LONG" -> {
                String chunk = "the quick brown fox jumps over the lazy dog ";
                while (sb.length() < 512) {
                    sb.append(chunk);
                }
            }
            case "SPARSE_ESCAPES" -> {
                String chunk = "value-without-specials-";
                for (int i = 0; i < 25; i++) {
                    sb.append(chunk);
                    if (i % 4 == 0) {
                        sb.append('"').append('\n').append('\t');
                    }
                }
            }
            case "DENSE_ESCAPES" -> {
                for (int i = 0; i < 300; i++) {
                    sb.append(i % 3 == 0 ? "x" : (i % 3 == 1 ? "\n" : "\""));
                }
            }
            case "ALL_ESCAPES" -> {
                for (int i = 0; i < 300; i++) {
                    sb.append('\n');
                }
            }
            default -> throw new IllegalArgumentException("Unknown profile: " + profile);
        }
        return sb.toString();
    }

    // -----------------------------------------------------------------
    // Legacy implementation, copied verbatim from JSONWriter#escape0
    // -----------------------------------------------------------------

    private static void legacyEscape(Writer writer, char[] characters) throws IOException {
        StringBuilder buffer = new StringBuilder();

        buffer.append('"');
        for (char c : characters) {
            switch (c) {
                case '"':
                    buffer.append("\\\"");
                    break;
                case '\\':
                    buffer.append("\\\\");
                    break;
                case '\b':
                    buffer.append("\\b");
                    break;
                case '\f':
                    buffer.append("\\f");
                    break;
                case '\n':
                    buffer.append("\\n");
                    break;
                case '\r':
                    buffer.append("\\r");
                    break;
                case '\t':
                    buffer.append("\\t");
                    break;
                default:
                    if (c <= 0x1F) {
                        buffer.append("\\u00");
                        buffer.append((char) (((c & 0xf0) >> 4) + '0'));
                        buffer.append((char) ((c & 0x0f) + '0'));
                    } else {
                        buffer.append(c);
                    }
                    break;
            }
        }
        buffer.append('"');

        writer.write(buffer.toString());
    }

    // -----------------------------------------------------------------
    // Optimized implementation, copied verbatim from JSONWriter#escape(String)
    // and its writeEscape() helper
    // -----------------------------------------------------------------

    private static final char[] HEX_CHARS = "0123456789abcdef".toCharArray();


// Precomputed lookup table for ASCII characters needing escape sequences
    private static final String[] ESCAPE_LUT = new String[128];
    static {
        // Control characters (0x00 to 0x1F)
        for (int i = 0; i < 32; i++) {
            ESCAPE_LUT[i] = String.format("\\u%04x", i);
        }
        // Specific JSON defined escape sequences
        ESCAPE_LUT['\b'] = "\\b";
        ESCAPE_LUT['\t'] = "\\t";
        ESCAPE_LUT['\n'] = "\\n";
        ESCAPE_LUT['\f'] = "\\f";
        ESCAPE_LUT['\r'] = "\\r";
        ESCAPE_LUT['"']  = "\\\"";
        ESCAPE_LUT['\\'] = "\\\\";
    }

private static void optimizedEscape(Writer writer, String str) throws IOException {
        writer.write('"');
        
        int length = str.length();
        int runStart = 0;

        for (int i = 0; i < length; i++) {
            char c = str.charAt(i);
            
            // Fast-path for non-escaped ASCII and high-unicode characters
            if (c >= 128) {
                continue;
            }
            
            String escaped = ESCAPE_LUT[c];
            if (escaped == null) {
                continue;
            }

            // Flush the unescaped segment accumulator
            if (i > runStart) {
                writer.write(str, runStart, i - runStart);
            }

            writer.write(escaped);
            runStart = i + 1;
        }

        // Flush any remaining characters
        if (runStart < length) {
            writer.write(str, runStart, length - runStart);
        }

        writer.write('"');
    }
    private static void optimizedEscape(Writer writer, char c) throws IOException {
        writer.write('"');
        if (c > 0x1F && c != '"' && c != '\\') {
            writer.write(c);
        } else {
            writeEscape(writer, c);
        }
        writer.write('"');
    }


    private static void writeEscape(Writer writer, char c) throws IOException {
        switch (c) {
            case '"':  writer.write("\\\""); break;
            case '\\': writer.write("\\\\"); break;
            case '\b': writer.write("\\b"); break;
            case '\f': writer.write("\\f"); break;
            case '\n': writer.write("\\n"); break;
            case '\r': writer.write("\\r"); break;
            case '\t': writer.write("\\t"); break;
            default:
                writer.write("\\u00");
                writer.write(HEX_CHARS[(c >>> 4) & 0x0F]);
                writer.write(HEX_CHARS[c & 0x0F]);
                break;
        }
    }

    // -----------------------------------------------------------------
    // Benchmarks
    // -----------------------------------------------------------------

    /**
     * Legacy path exactly as it would really be invoked: a String value has to
     * be converted to a char[] before escape0() can run, so that conversion
     * cost is deliberately included here. This is the fair, apples-to-apples
     * comparison against optimizedEscape_fromString().
     */


    
    @Benchmark
    public long legacyEscape0_fromString() throws IOException {
        legacyEscape(writer, testString.toCharArray());
        return writer.count;
    }
 
    /**
     * Legacy path when a char[] is already on hand (no conversion cost) -
     * a lower bound for escape0's own algorithmic cost in isolation.
     */
    
/**
     * Measures Legacy implementation handling single characters by wrapping 
     * them in a temporary char[] array iteratively.
     */
    @Benchmark
    public long legacyEscape0_singleChars() throws IOException {
        char[] chars = testChars;
        for (int i = 0; i < chars.length; i++) {
            // This mirrors Impl 2: value instanceof Character -> escape(writer, new char[]{(char)value})
            legacyEscape(writer, new char[] { chars[i] });
        }
        return writer.count;
    }
 
    /**
     * Optimized path operating directly on the String, no intermediate array.
     */


    
    @Benchmark
    public long optimizedEscape_fromString() throws IOException {
        optimizedEscape(writer, testString);
        return writer.count;
    }



    /**
     * Measures Optimized implementation handling single characters directly
     * without any array allocation or stream buffering.
     */
    @Benchmark
    public long optimizedEscape_singleChars() throws IOException {
        char[] chars = testChars;
        for (int i = 0; i < chars.length; i++) {
            // This mirrors Impl 1: value instanceof Character -> escape(writer, c)
            optimizedEscape(writer, chars[i]);
        }
        return writer.count;
    }
    // -----------------------------------------------------------------
    // Sink Writer: counts characters instead of actually storing/copying
    // them, so the benchmark measures the escaping algorithm itself rather
    // than downstream I/O or buffer-growth overhead.
    //
    // Overriding write(String, int, int) matters: without it,
    // java.io.Writer's default implementation allocates and copies a char[]
    // anyway, which would silently defeat the "zero-allocation" benefit
    // being measured for optimizedEscape().
    // -----------------------------------------------------------------

    static final class CountingWriter extends Writer {
        long count;

        @Override
        public void write(int c) {
            count++;
        }

        @Override
        public void write(char[] cbuf, int off, int len) {
            count += len;
        }

        @Override
        public void write(String str, int off, int len) {
            count += len;
        }

        @Override
        public void flush() {
            // no-op
        }

        @Override
        public void close() {
            // no-op
        }
    }
}
```

grgrzybek:
Interesting... Let me digest it first ;)

grgrzybek:
Enhanced CORS is implemented:
* #143 

I want to add support for proxy headers for restrictions:
* #621

And together with this PR I can release 2.6.1 next week (hopefully!)

grgrzybek:
Thank you very much @GChuf for optimising the serializer!
And for the `0x1A` fix!

Issue #1029, opened 2026-07-10
Improve and document current single-thread JVM Agent usage


grgrzybek:
Here's the reason:
* #1026
* https://github.com/jolokia/jolokia/commit/6069dc47cfa735befcdfe8e8951a248908748c4c
* https://docs.oracle.com/en/java/javase/21/docs/api/jdk.httpserver/module-summary.html

This AI generated PR is simply an indication that Jolokia does not set `sun.net.httpserver.maxReqTime` system property. And it shouldn't, because who knows what other parts of JVM rely on built-in JDK HTTP Server?

But this doesn't mean we don't need to do anything:
* `single` is not a good default for HTTP Server executor
* Jolokia should reconsider sending `Connection: close` response header - because when we _navigate_ to Jolokia URL from a browser, the browser usually sends `Connection: keep-alive` and it'll also keep the connection awaiting for next HTTP request over existing TCP connection by re-adding the task to a queue. It is always a bad practice when single-threaded executor is used.
* Sending partial requests (as shown in #1026) always keeps the single thread busy and no other client can connect.

Technically, when JDK HTTP Server gets `Connection: keep-alive`, it may mark the connection as "let's close it" when the response header contains `Connection: close`, but when it doesn't do this, we have:

1. `sun.net.httpserver.ServerImpl.Dispatcher#handleEvent()` adds existing connection to `sun.net.httpserver.ServerImpl.Dispatcher#connsToRegister` in `HTTP-Dispatcher` thread
2. `sun.net.httpserver.ServerImpl.Dispatcher#reRegister()` is called for this connection and it's added to `sun.net.httpserver.ServerImpl#idleConnections()`
3. `java.nio.channels.Selector#select(long)` continues and selected keys is not empty. Our not closed connection is "read ready"
4. `sun.net.httpserver.ServerImpl#requestStarted()` is called on this idle connection and it's added to `sun.net.httpserver.ServerImpl#reqConnections`
5. `sun.net.httpserver.ServerImpl.Dispatcher#handle()` is called which executes an `sun.net.httpserver.ServerImpl.Exchange` task in the executor
6. single thread `jolokia-0` receives a task from the queue in `java.util.concurrent.ThreadPoolExecutor#getTask()`
7. `sun.net.httpserver.ServerImpl.Exchange#run()` is called and `sun.net.httpserver.Request#Request()` constructor tries to read new HTTP request hanging at:
```
[...]

PR #1030, opened 2026-07-13
Use application/json instead of text/json content-type header

GChuf:
Noticed this one while walking around ...
JSON payloads should be served with application/json mimetype.
https://www.ietf.org/rfc/rfc4627.txt


grgrzybek:
I remember there were _some_ issues with this and webpack-dev-server and/or tests. I'll get back to it after holidays (end of July).

This suprise prevented me from releasing 2.6.1 earlier:
* #1029

GChuf:
Hmm - I would expect text/json to be more problematic, if anything. Although my reverse proxies don't seem to mind (responses still gzipped even though I did not specify text/json mimetype).
Do have a nice time and don't think about headers and HTTP requersts until the end of July! ...

grgrzybek (approved):
Thanks @GChuf!
Now it's perfectly consistent (Hawtio also uses `application/json`).

